PR #687, opened 2023-12-11
Expanded examples docstring for regridder

Manangka:
In GitLab by @jc-hunink on Nov 25, 2021, 14:30

Small changes in the documentation of class Regridder(object)

### DIFF
--- a/imod/prepare/regrid.py
+++ b/imod/prepare/regrid.py
@@ -346,6 +346,11 @@ class Regridder(object):
 
     >>> result = imod.prepare.Regridder(method="mean").regrid(source, like)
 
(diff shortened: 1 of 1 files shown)


Manangka:
In GitLab by @jc-hunink on Nov 25, 2021, 21:10

requested review from @Huite

Manangka:
In GitLab by @JoerivanEngelen on May 11, 2022, 17:20

@Huite Could you verify if the docstring added is true? It probably requires some clarification.

I see the ``regrid`` method can regrid along 3 dimensions, so I guess ``("z", "y", "x")``, right? Or does that include ``("time", "y", "x")`` as well? We could point to ``LayerRegridder`` here for people who have a ``layer`` dimension.

Issue #553, opened 2023-09-26
Fix issue with wells in inactive cells while splitting domain


Manangka:
Run the ex01_twri_partioned.py with the following submodel array:

```plaintext
gwf_model = simulation["GWF_1"]
active = gwf_model.domain.sel(layer=1)

submodel_labels = xr.ones_like(active).where((active.x > 40000) & (active.y < 40000), 0)
```

```plaintext
Return the following error:
ERROR REPORT:

   1.   Cell is outside active grid domain: (1,11,14)
   2.   Cell is outside active grid domain: (1,13,10)
   3.   Cell is outside active grid domain: (1,13,12)
   4.   Cell is outside active grid domain: (1,13,14)
   5.   Cell is outside active grid domain: (1,9,10)
   6.   Cell is outside active grid domain: (1,9,12)
   7.   Cell is outside active grid domain: (1,9,14)
   8.   Cell is outside active grid domain: (1,11,10)
   9.   Cell is outside active grid domain: (1,11,12)
  10.            9  errors encountered.

UNIT ERROR REPORT:

  1. Error occurred while reading file
     'C:\Users\titus\AppData\Local\Temp\tmp68v0ffbf\GWF_1_0\wel\wel.dat'

```

desired solution: wells should be removed from the high level well package if in inactive cells while splitting the model ( or by clipping in general) 
we think this is only an issue  for structured grids.

Manangka:
added #538 as parent issue

Manangka:
removed parent issue #538

Manangka:
changed type from task to incident

Manangka:
In GitLab by @JoerivanEngelen on Dec 6, 2023, 16:58

marked this incident as related to #680

PR #693, opened 2023-12-11
Issue #680 "Partitioning doesn't work for Wells"

### LINKED ISSUE #680: Partitioning doesn't work for Wells
In GitLab by @JoerivanEngelen on Dec 6, 2023, 16:45

Partitioning doesn't work for Wells, because the ``clip_by_grid`` method doesn't work for IPointData (Wells):

* The clipping method is based on the assumption that ``active`` is already the clipped grid, which it isn't. It appears to be a mask.
* There is no implementation for StructuredGrids

We apparently also lack serious testing for Wells in this regard.

Looking at the code, I think it should be relatively easy to use ``imod.select.points_values`` to throw wells outside grid, it also allows easy support of StructuredGrids.

This will potentially fix this as well: https://gitlab.com/deltares/imod/imod-python/-/issues/553

Manangka:
In GitLab by @JoerivanEngelen on Dec 11, 2023, 13:18

Closes #680, #700 and #553

This PR fixes the following:
- Change behaviour Well.mask method so it masks in 2D grids instead deepcopying
- clip_by_grid should always operate on 2d grids, correct tests for that
- refactor test_mf6_partitioning to use pytest cases instead of pytest parametrization
- fix erronous default value for ``out_of_bounds`` in ``point_values``
- fix bug where well could not be assigned to the first cell in an unstructured grid

### DIFF
--- a/imod/mf6/utilities/clip.py
+++ b/imod/mf6/utilities/clip.py
@@ -11,6 +11,7 @@ from imod.mf6.interfaces.ipackagebase import IPackageBase
 from imod.mf6.interfaces.ipointdatapackage import IPointDataPackage
 from imod.mf6.utilities.dataset import get_scalar_variables
 from imod.mf6.utilities.grid import get_active_domain_slice
+from imod.select.points import points_values
 from imod.typing import GridDataArray, ScalarDataset
 from imod.typing.grid import bounding_polygon, is_spatial_2D
 
@@ -51,12 +52,13 @@ def clip_by_grid(package: IPackageBase, active: xu.UgridDataArray) -> IPackageBa
 
 @typedispatch
 def clip_by_grid(
-    package: IPointDataPackage, active: xu.UgridDataArray
+    package: IPointDataPackage, active: GridDataArray
 ) -> IPointDataPackage:
-    """Clip PointDataPackage outside unstructured grid."""
-    points = np.column_stack((package.x, package.y))
+    """Clip PointDataPackage outside grid."""
 
-    is_inside_exterior = active.grid.locate_points(points) != -1
+    point_active = points_values(active, x=package.x, y=package.y)
+
+    is_inside_exterior = point_active == 1
     selection = package.dataset.loc[{"index": is_inside_exterior}]
 
     cls = type(package)
--- a/imod/mf6/utilities/clip.py
+++ b/imod/mf6/utilities/clip.py
@@ -11,7 +11,6 @@ from imod.mf6.interfaces.ipackagebase import IPackageBase
 from imod.mf6.interfaces.ipointdatapackage import IPointDataPackage
 from imod.mf6.utilities.dataset import get_scalar_variables
 from imod.mf6.utilities.grid import get_active_domain_slice
-from imod.select.points import points_values
 from imod.typing import GridDataArray, ScalarDataset
 from imod.typing.grid import bounding_polygon, is_spatial_2D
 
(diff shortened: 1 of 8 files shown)


JoerivanEngelen:
Things I encountered:
We are inconsistent with accepted dimensionalities for the ``clip_by_grid`` and ``mask`` methods: Sometimes 3D, sometimes 2D. Probably good to pick that up in a separate issue.

Manangka:
Could you Prepend "Issue #680" to the PR title? That way it will also end up in the git commit message

imod/mf6/utilities/clip.py
```diff
@@ -54,9 +54,13 @@ def clip_by_grid(
     package: IPointDataPackage, active: xu.UgridDataArray
 ) -> IPointDataPackage:
     """Clip PointDataPackage outside unstructured grid."""
+
+    domain_slice = get_active_domain_slice(active)
+    active_clipped = active.isel(domain_slice, missing_dims="ignore")
```
Manangka, line 59: Why is this necessary?  Wouldn't this just return active?

imod/mf6/wel.py
```diff
@@ -494,18 +492,32 @@ def regrid_like(self, target_grid: GridDataArray, *_) -> "Well":
         grid-agnostic, instead this method clips the package based on the grid
         exterior.
         """
-        return clip_by_grid(self, target_grid)
+        target_grid_2d = target_grid.isel(layer=0, drop=True, missing_dims="ignore")
+        return clip_by_grid(self, target_grid_2d)
 
-    def mask(self, _) -> Package:
+    def mask(self, domain: GridDataArray) -> "Well":
```
Manangka, line 498: If you add `from __future__ import annotations` then you can add the Well type hint as a regular hint instead of a string: `-> Well`

imod/mf6/wel.py
```diff
@@ -494,18 +492,32 @@ def regrid_like(self, target_grid: GridDataArray, *_) -> "Well":
...
+        # Drop layer coordinate if present, otherwise a layer coordinate is assigned
+        # which causes conflicts downstream when assigning wells and deriving
+        # cellids.
+        domain = domain.isel(layer=0, drop=True, missing_dims="ignore").drop(
+            "layer", errors="ignore"
+        )
+        point_active = points_values(domain, x=self.x, y=self.y)
+
+        is_inside_exterior = point_active == 1
+        selection = self.dataset.loc[{"index": is_inside_exterior}]
```
Manangka, line 515: The above lines is the same as the clip_by_grid method. Can't you call that method to reduce code duplication?

imod/select/points.py
```diff
@@ -243,7 +243,7 @@ def points_indices(da, out_of_bounds="raise", **points):
     return indices
 
 
-def points_values(da, out_of_bounds="error", **points):
+def points_values(da, out_of_bounds="raise", **points):
```
Manangka, line 246: Why is this needed?

imod/tests/test_mf6/test_multimodel/test_mf6_partitioning_unstructured.py
```diff
@@ -206,3 +206,63 @@ def test_partitioning_unstructured_with_vpt_cells(
...
+)
+def test_partitioning_unstructured_with_well(
+    tmp_path: Path, circle_model: Modflow6Simulation, partition_name: str
+):
+    simulation = circle_model
+    # increase the recharge to make the head gradient more pronounced
+    simulation["GWF_1"]["rch"]["rate"] *= 100
+
+    # Add well
+    well = imod.mf6.Well(
```
luitjansl, line 224: why do you add just 1 well? Isn't that a bit meager for testing this functionality?

imod/mf6/utilities/clip.py
```diff
@@ -54,9 +54,13 @@ def clip_by_grid(
     package: IPointDataPackage, active: xu.UgridDataArray
 ) -> IPointDataPackage:
     """Clip PointDataPackage outside unstructured grid."""
+
```
luitjansl, line 57: in many tests, "active" is taken as the top layer of idomain; and the top layer of idomain may sometimes contain some vertical passthrough (vpt) cells, even if the cells below it are active. Should wells coinciding with vpt cells stay or go? Or should we raise an exception if active contains vpt cells? Or should we make a function to infer active from idomain, and only assign vpt if all layers are vpt in idomain for a particular cell?

luitjansl (commented):
looks good, some comments

imod/mf6/utilities/clip.py
JoerivanEngelen: The way we select data in the partitioning is with a 2D grid of inactive and active cells. The line:

```is_inside_exterior = active_clipped.grid.locate_points(points) != -1```

only checks if wells are located within the grid exterior. So inactive cells have to be removed from the grid.

imod/tests/test_mf6/test_multimodel/test_mf6_partitioning_unstructured.py
JoerivanEngelen: Good point, added this. Immediately found a bug and fixed it! https://github.com/Deltares/imod-python/issues/700

imod/mf6/utilities/clip.py
JoerivanEngelen: Good point. I think we need to decide what should be 3D and what should be 2D, I'll create a separate issue for this.

luitjansl (approved):

Issue #704, opened 2023-12-20
Switch pooch download location to the new GitHub location


Manangka:
As part of the GitHub migration the https://gitlab.com/deltares/imod/imod-artifacts has been moved to https://github.com/Deltares/imod-artifacts.

PR #705, opened 2023-12-20
Issue #704 Update the artifacts download location

Manangka:
As part of the GitHub migration the https://gitlab.com/deltares/imod/imod-artifacts has been moved to https://github.com/Deltares/imod-artifacts.

This commit updates the artifacts location. Furthermore the datasets.py file has been removed as it seems to be unused

### DIFF
--- a/imod/data/sample_data.py
+++ b/imod/data/sample_data.py
@@ -22,7 +22,7 @@ except ImportError:
 
 REGISTRY = pooch.create(
     path=pooch.os_cache("imod"),
-    base_url="https://gitlab.com/deltares/imod/imod-artifacts/-/raw/main/",
+    base_url="https://github.com/Deltares/imod-artifacts/raw/main/",
     version=None,
     version_dev="main",
     env="IMOD_DATA_DIR",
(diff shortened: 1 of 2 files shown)


JoerivanEngelen:
Thanks for migrating this!

Your removal of example_data.py got me thinking:
I think there still is a place for the example_data.py (although unused). We could put some example models in here, created dynamically by iMOD Python. We could put these examples in example_data.py: https://github.com/Deltares/imod-python/blob/master/examples/mf6/example_models.py

These example models can then easily be called in the test suite easily as well, possibly also removing the need for test fixtures like: https://github.com/Deltares/imod-python/blob/master/imod/tests/fixtures/mf6_twri_fixture.py

Manangka:
> Thanks for migrating this!
> 
> Your removal of example_data.py got me thinking: I think there still is a place for the example_data.py (although unused at the moment). We could put some example models in here, created dynamically by iMOD Python. We could put these examples in example_data.py: https://github.com/Deltares/imod-python/blob/master/examples/mf6/example_models.py
> 
> These example models can then easily be called in the test suite easily as well, possibly also removing the need for test fixtures like: https://github.com/Deltares/imod-python/blob/master/imod/tests/fixtures/mf6_twri_fixture.py

For some of the examples this is already happening. E.g. The Hondsrug and the circle examples use data retrieved from this blob

JoerivanEngelen (approved):

Issue #560, opened 2023-10-04
Remove setup.cfg


Manangka:
All metadata can moved to the pyproject.toml file and the setup.cfg file can be removed
https://packaging.python.org/en/latest/specifications/declaring-project-metadata/#declaring-project-metadata

The setup.cfg file is an artefact of the setuptools module. Nowadays all the information stored inside the setup.cfg can also be stored in the pyproject.toml. This has the advantage that all the project information is stored in 1 location

PR #707, opened 2023-12-21
Issue #560 Remove setup.cfg

Manangka:
In this commit all the contents of setup.cfg has been moved to the pyproject.toml file.
Flake8 doesn't support the pyproject.toml format, therefor it has been relaced with ruff.


Huite:
I would recommend moving to ruff instead and totally getting rid of flake.

It's likely that ruff is a bit stricter on a number of fields. I've been using the config that I think @Hofer-Julian set up for Ribasim with good effect: 

See e.g.:
https://github.com/Deltares/xugrid/blob/main/ruff.toml

Manangka:
Alright, I will have a look at Ruff

Hofer-Julian:
Ruff simply reimplements flake8, isort, black and many others. But it's much faster and you only need to install one thing

pyproject.toml
```diff
@@ -2,13 +2,137 @@
...
+    "Development Status :: 4 - Beta",
+    "Intended Audience :: Science/Research",
+    "Topic :: Scientific/Engineering :: Hydrology",
+    "License :: OSI Approved :: MIT License",
+    "Operating System :: OS Independent",
+    "Programming Language :: Python",
+    "Programming Language :: Python :: 3",
+    "Programming Language :: Python :: 3.8",
+    "Programming Language :: Python :: 3.9",
+    "Programming Language :: Python :: 3.10",
```
JoerivanEngelen, line 29: This is outdated info, should be:

```suggestion
    "Programming Language :: Python :: 3.10",
    "Programming Language :: Python :: 3.11",
```

pyproject.toml
```diff
@@ -2,13 +2,137 @@
...
+    "tqdm",
+    "xarray >=0.15",
+    "xugrid >=0.3.0",
+]
+
+[project.urls]
+Homepage = "https://github.com/Deltares/imod-python"
+Documentation = "https://deltares.github.io/imod-python/"
+Repository = "https://github.com/Deltares/imod-python.git"
+"Bug Tracker" = "https://github.com/Deltares/imod-python/issues"
```
JoerivanEngelen, line 56: Why the quotes around "Bug Tracker"?

JoerivanEngelen (changes requested):

pyproject.toml
Manangka: I followed the example on:
https://packaging.python.org/en/latest/guides/writing-pyproject-toml/#advanced-plugins

In that example the "Bug Tracker" is also in quotes

JoerivanEngelen (approved):

Issue #659, opened 2023-12-11
Test and fix (if necessary) partitioning for HFB package


Manangka:
In GitLab by @JoerivanEngelen on Nov 22, 2023, 12:37

JoerivanEngelen:
Found the following issues: https://github.com/Deltares/imod-python/issues/696, https://github.com/Deltares/imod-python/issues/697

PR #710, opened 2023-12-22
Issue #709 fix hfb structured grids

### LINKED ISSUE #709: Issues with HFB lines not properly assigned to structured grids
Adding tests for HFBs with structured grids raised the following issues:

* HorizontalFlowBarrier lines are currently not properly clipped in clip_by_grid, due to bounding_polygon also returning a polygon for active values.
* When the barrier is placed at the edge of a model partition it is removed. This can create empty HFB packages which should be removed, this should raise an error.
* The ``clip_by_grid`` can introduce MultiPoints and MultiLineStrings for ILineDataPackage. The MultiPoints need to be thrown out, theMultiLineStrings can be separated into separate LineStrings.

JoerivanEngelen:
Closes #709 and #659

This PR implements a fix for:

* HorizontalFlowBarrier lines not being properly clipped in clip_by_grid, due to bounding_polygon also returning a polygon for active values.
* When the barrier is placed at the edge of a model partition it is removed. This can create empty HFB packages which should be removed, this now raises an error.
* The clip_by_grid can introduce MultiPoints and MultiLineStrings for ILineDataPackage. The MultiPoints are thrown out, the MultiLineStrings are separated into separate LineStrings.

UPDATE 03/01/2024:
after comment https://github.com/Deltares/imod-python/pull/710/files/4df2e20c82b07c111890cd85d85696924da3296b#r1439526060
some code was refactored into a separate property ``line_data``. This has the advantage that it is now more accessible to access line data as, and set it with, GeoDataFrames.


imod/mf6/utilities/clip.py
```diff
@@ -117,9 +117,14 @@ def clip_by_grid(package: ILineDataPackage, active: GridDataArray) -> ILineDataP
...
-    # (type_id == 0) is returned. Drop these.
-    is_points = shapely.get_type_id(package_gdf_clipped.geometry) == 0
+    # or multipoint is returned. Drop these.
+    type_ids = shapely.get_type_id(package_gdf_clipped.geometry)
+    is_points = (type_ids == shapely.GeometryType.POINT) | (
+        type_ids == shapely.GeometryType.MULTIPOINT
+    )
     package_gdf_clipped = package_gdf_clipped[~is_points]
+    # Separate MultiLinestrings ino separate Linestrings
+    package_gdf_clipped = package_gdf_clipped.explode("geometry", ignore_index=True)
```
Manangka, line 127: This method is calling methods that are not on the ILineDataPackage interface but on the implementation of that interface. Therefor it is violating the Liskov substitution principle.  To avoid this i have altered this methods slightly:

```
@typedispatch
def clip_by_grid(package: ILineDataPackage, active: GridDataArray) -> ILineDataPackage:
    """Clip LineDataPackage outside unstructured/structured grid."""

    # Clip line with polygon
    bounding_gdf = bounding_polygon(active)
    clipped_geometry = package.geometry.clip(bounding_gdf)

    # Catch edge case: when line crosses only vertex of polygon, a point
    # or multipoint is returned. Drop these.
    type_ids = shapely.get_type_id(clipped_geometry.geometry)
    is_points = (type_ids == shapely.GeometryType.POINT) | (
        type_ids == shapely.GeometryType.MULTIPOINT
    )
    clipped_geometry = clipped_geometry[~is_points]

    # Convert MultiLineStrings to LineStrings
    clipped_geometry = clipped_geometry.transform({'geometry': lambda x: x if shapely.get_type_id(x) == 1 else shapely.ops.linemerge(x)})

    # Create new instance
    clipped_package = deepcopy(package)
    clipped_package.geometry = clipped_geometry
    return clipped_package
```

This has a benefit that the _getsettings, _get_variables_names_for_gdf and the _line_package_to_gdf methods are no longer needed and can be removed.


2 more changes are needed.  The ILineDataPackage becomes:
```
class ILineDataPackage(IPackageBase):
    """
    Interface for packages for which the data is defined by lines independent of the domain definition.
    """
    @property
    @abstractmethod
    def geometry(self) -> gpd.GeoDataFrame:
        raise NotImplementedError

    @geometry.setter
    @abstractmethod
    def geometry(self, value: gpd.GeoDataFrame) -> None:
        raise NotImplementedError
```

and the implementations for those methods in the hfb package are:
```
    @property
    def geometry(self) -> gpd.GeoDataFrame:
        return gpd.GeoDataFrame(geometry=self.dataset["geometry"].values)

    @geometry.setter
    def geometry(self, value: gpd.GeoDataFrame) -> None:
        self.dataset["geometry"] = value.to_xarray()["geometry"]
        self.dataset = self.dataset.dropna(dim="index")
```

Manangka (changes requested):

imod/mf6/hfb.py
```diff
@@ -281,11 +282,28 @@ def __init__(
...
+        ] + self._get_vertical_variables()
 
     @property
-    def geometry(self) -> np.ndarray[object]:
-        return self.dataset["geometry"].values
+    def line_data(self) -> gpd.GeoDataFrame:
+        variables_for_gdf = self._get_variable_names_for_gdf()
+        return gpd.GeoDataFrame(
+            self.dataset[variables_for_gdf].to_dataframe(),
+            geometry="geometry",
```
Manangka, line 298: I don't think we need the `geometry="geometry" argument. I think because there is already a variable called geometry in the dataset that will automatically be used as the geometry column.

Manangka (approved):

imod/mf6/hfb.py
JoerivanEngelen: I think it is better to specify that explitictly to geopandas, as the geometry column is crucial to its functioning. On one hand it might be that the argument name "geometry" changes, on the other hand it might be that behavior changes for inferring this "geometry" column if no geometry is explicitly provided. I prefer the first case, as it provides a clearer erro message.

Issue #352, opened 2023-04-26
Support pandas 2.0


Manangka:
In GitLab by @JoerivanEngelen on Apr 26, 2023, 11:33

Things to pay attention to:
- Xarray should be pinned higher than this release: https://github.com/pydata/xarray/releases/tag/v2023.04.0
- I remember reading things changed with the time handling, we test investigate if our cftime things still work

Manangka:
In GitLab by @JoerivanEngelen on Dec 4, 2023, 14:41

marked this issue as related to #676

PR #715, opened 2024-01-03
Issue #352 Support pandas 2

JoerivanEngelen:
This PR makes iMOD Python support pandas 2 (or at least our CI).

Fixes: https://github.com/Deltares/imod-python/issues/352

Pandas 2 introduced some breaking changes, especially how time is handled. Updating to Pandas 2 resulted in three tests breaking. This PR fixes the following issues after the update:

- Apparently resetting the index after a groupby, the index is not dropped with pandas 2, this needs to be set explictly.
- [The read_csv function deprecated inferring datetime](https://pandas.pydata.org/pandas-docs/stable/reference/api/pandas.read_csv.html#pandas.read_csv). Not doing so breaks the ``point_interpolation.py`` example. Conversion to datetime has to be done manually with a call to [pd.to_datetime](https://pandas.pydata.org/pandas-docs/stable/reference/api/pandas.to_datetime.html#pandas.to_datetime).
- Set ``use_absolute_paths`` flag in ``test_mf6_flopy_compatibility.py`` to True. Which was the original intention of the test. It appears relative paths accidentily worked for Flopy 3.4, but don't work anymore for Flopy 3.5. With pandas 2, Flopy 3.5 can be installed, which introduced this error.


Manangka (approved):

JoerivanEngelen:
A test is failing for the flopy compatibility. I could reproduce this locally with using the 3.5.0 flopy version, with the 3.4.1 version everything works. Not sure if this is related to this branch.

JoerivanEngelen:
I managed to fix this locally, there seemed to be an error in the way we parametrized ``absolute_paths``. It was set to False (creating relative paths), instead of using absolute paths. We implemented the ``use_absolute_paths`` feature just to work around these Flopy errors, so they should be True.

It appears relative paths accidentily worked for Flopy 3.4, but don't work anymore for Flopy 3.5

Issue #718, opened 2024-01-04
Add separate tasks for the unit tests and the examples in pixi


Manangka:
To run only a subset of the test suite the test task needs to be split. A natural split is the one between running the examples and the unit tests.

This split will also enables us to run the tests in parallel on the CI servers

PR #719, opened 2024-01-04
Issue #718 Create separate tasks for running the unit tests and the examples

Manangka:
This commit splits the test task in a unit test task and an examples task.
This will allow us to run only a subset of the test suite. It will also allows to run the test in parallel on the CI servers


luitjansl (approved):
looks good.

JoerivanEngelen (approved):

Issue #708, opened 2023-12-21
Use an exact join to initiate Package datasets


JoerivanEngelen:
We currently initialize the package dataset by first instantiating a ``xr.Dataset`` or ``xu.UgridDataset`` and then updating this each time by assigning new DataArrays to it. This has the disadvantage that we do not have control over how coordinates should be joined, as the default value for ``join`` is used, which is ``"outer"`` :

[Call to merge_core, no option for join](https://github.com/pydata/xarray/blob/main/xarray/core/merge.py#L1052)

[Definition merge_core with default value](https://github.com/pydata/xarray/blob/main/xarray/core/merge.py#L646)

iMOD Python expects its dataset to have exactly joined coordinates. To have more control over this, we should consider initializing package datasets with ``xr.merge``, which allows us to set the ``join`` method to ``exact``. This will the following error: 

```
ValueError: cannot align objects with join='exact' where index/labels/sizes 
are not equal along these coordinates (dimensions): 'layer' ('layer',)
```

It's not entirely perfect, as the name of the erronous variable is not printed (we do have ``idomain``, so we know what the coordinates should look like). But it ensures that the thing we implicitly expect as a dataset is created correctly. Otherwise we run into issues like: 

https://github.com/Deltares/imod-python/issues/674

UPDATE (03/01/2024):
As discussed with @Manangka 

* The call to ``xr.merge`` creates a Dataset and handles settings (scalar variables) gracefully. This should be the first thing done upon package initialization.
* We can create a ``imod.merge`` function which typedispatches UgridDataArray to work around issue https://github.com/Deltares/xugrid/issues/179. But better to resolve this in ``xugrid`` upstream.
* We should avoid copies of DataArrays are made senselessly. ``to_dataset`` might do this (but xarray also call this behind the scenes when assigning a DataArray to a dataset, so might not?)

JoerivanEngelen:
This might be a blocker to create UgridDatasets: https://github.com/Deltares/xugrid/issues/179, but we might be able to work around it.

JoerivanEngelen:
Just as an extra check to see if xr.merge causes any loss of performance in this case as reported in: https://github.com/Deltares/imod-python/issues/736#issuecomment-1884643780 . Apparently with nicely overlapping coordinates, merging is not such a slow process and the changes made in this PR should only (very slightly) enhance performance.

```python
# %% import
import numpy as np
import xarray as xr


# %% functions
def make_da(nlay, nrow, ncol):
[...]

Issue #674, opened 2023-11-30
The SpecificStorage class returns the wrong error message for a mismatch in dims


Manangka:
In GitLab by @HendrikKok on Nov 30, 2023, 16:04

For the variable ‘convertible’, the error message is ‘float64 != np.int’ is raised, while:

1-	dtype of array is np.int32,
2-	the given array had shape of (nlayer – 1).

Probably the correct error on the array shape, returns the wrong error message.

Manangka:
In GitLab by @HendrikKok on Dec 5, 2023, 11:03

Hi Joeri,

I tested it on a large model (as always). The quickest way to reproduce this, is to use the existing tests for the sto-package I think. For the convertible argument use a 3d xarray with the correct dtype, but with a layer dimension of size layer -1. This should result in this error.

Manangka:
In GitLab by @HendrikKok on Dec 5, 2023, 13:47

at voila @ test-error-message-sto-package-mf6

JoerivanEngelen:
Thanks! I checked the branch and I think this stems of a wider inconvenience: The DataArrays provided to a Package are immediately set as variable in a dataset upon Package initialization. Then a similar thing happens as calling:

```
xr.merge([{"specific_yield": sy_layered, "transient": transient, "convertible": convertible_missing_layer}], join="outer")
```

This will replace the missing indexes in ``convertible_missing_layer`` with ``np.nan``, which causes the integer array to be converted to float. In that sense the validation is doing what it is supposed to do: It detects a coordinate with the wrong type in your dataset.

We could make this more explicit, because iMOD Python expects coordinates to match exactly. We could therefore take an alternative approach to initiating the dataset and use something like:

```
xr.merge([{"specific_yield": sy_layered, "transient": transient, "convertible": convertible_missing_layer}], join="exact")
```

This throws the error:

```
ValueError: cannot align objects with join='exact' where index/labels/sizes are not equal along these coordinates (dimensions): 'layer' ('layer',)
```

Which already points the user better to the culprit.

An alternative would be to implement validation schemes which should be called upon initialization BEFORE the dataset is created. This has the benefit that we could provide more info to the user, but the disadvantage that we are creating duplicate logic. So far the strategy of iMOD Python has been to leverage ``xarray`` where possible and extend where needed.

JoerivanEngelen:
Closing this for now (the culprit has been found), progress on resolving this issue can be followed here: https://github.com/Deltares/imod-python/issues/708

PR #722, opened 2024-01-05
Draft: Issue #708 Refactor dataset initialization

JoerivanEngelen:
The changeset is incomplete, as it is not rolled out for all packages yet. Therefore tests will fail, hence the draft status. This to make the review process more focused, reviewers are asked to provide feedback on the approach.

Changes:
- Variable names are inferred based on what is provided to locals(). This makes it possible to avoid having to manually assign variables in the __init__ of each specific package.
- I had to add one function to work around this issue: https://github.com/Deltares/xugrid/issues/179
- Grids are merged with an exact join, this to avoid that xarray joins variable coordinates in an unwanted way when dataarrays are inconsistent, which causes issues like https://github.com/Deltares/imod-python/issues/674
- Refactored the riv unit tests to use pytest cases, so that tests (without the xr.merge) are run for both unstructured and structured data.

### DIFF
--- a/imod/mf6/pkgbase.py
+++ b/imod/mf6/pkgbase.py
@@ -16,6 +16,37 @@ EXCHANGE_PACKAGES = "gwfgwf"
 ARGS_TO_EXCLUDE = ["validate"]
 
 
+def merge_unstructured_dataset(variables_to_merge):
+    """Work around xugrid issue https://github.com/Deltares/xugrid/issues/179"""
+
+    # Separate variables into list of grids and dict of scalar variables
+    grids_ls = []
+    scalar_dict = {}
+    for name, variable in variables_to_merge.items():
+        if isinstance(variable, xu.UgridDataArray):
+            grids_ls.append(variable.rename(name))
+        else:
+            scalar_dict[name] = variable
+
+    # Merge grids
+    dataset = xu.merge(grids_ls, join="exact")
+
+    # Assign scalar variables manually
+    for name, variable in scalar_dict.items():
+        dataset[name] = variable
+
+    return dataset
+
+
+def merge_at_init(variables_to_merge):
+    if variables_to_merge is not None:
+        for arg in variables_to_merge.values():
+            if isinstance(arg, xu.UgridDataArray):
+                return merge_unstructured_dataset(variables_to_merge)
+
+    return xr.merge([variables_to_merge], join="exact")
+
+
 class PackageBase(IPackageBase, abc.ABC):
     """
     This class is used for storing a collection of Xarray DataArrays or UgridDataArrays
@@ -28,15 +59,10 @@ class PackageBase(IPackageBase, abc.ABC):
     def __new__(cls, *_, **__):
         return super(PackageBase, cls).__new__(cls)
 
-    def __init__(self, allargs=None):
-        if allargs is not None:
-            for arg in allargs.values():
-                if isinstance(arg, xu.UgridDataArray):
-                    self.__dataset = xu.UgridDataset(grids=arg.ugrid.grid)
-                    return
-        self.__dataset = xr.Dataset()
+    def __init__(self, variables_to_merge=None):
+        self.__dataset = merge_at_init(variables_to_merge)
 
-    def _get_variable_names(init_method):
+    def _get_variable_names(self, init_method):
         """
         Return variable names based on the arguments provided to the classes'
         __init__ method. Removes argument names that need to be excluded.
--- a/imod/mf6/pkgbase.py
+++ b/imod/mf6/pkgbase.py
@@ -1,4 +1,5 @@
 import abc
+import inspect
 import numbers
 import pathlib
 
@@ -11,6 +12,7 @@ from imod.mf6.interfaces.ipackagebase import IPackageBase
 
 TRANSPORT_PACKAGES = ("adv", "dsp", "ssm", "mst", "ist", "src")
 EXCHANGE_PACKAGES = "gwfgwf"
+ARGS_TO_EXCLUDE = ["validate"]
 
 
 class PackageBase(IPackageBase, abc.ABC):
@@ -32,6 +34,17 @@ class PackageBase(IPackageBase, abc.ABC):
                     self.__dataset = xu.UgridDataset(grids=arg.ugrid.grid)
                     return
         self.__dataset = xr.Dataset()
+
+    def _get_variable_names(init_method):
+        """
+        Return variable names based on the arguments provided to the classes'
+        __init__ method. Removes argument names that need to be excluded.
+        """
+        variable_names = list(inspect.signature(init_method).parameters.keys())
+        for var_to_exclude in ARGS_TO_EXCLUDE:
+            if var_to_exclude in variable_names:
+                variable_names.remove(var_to_exclude)
+        return variable_names
 
     @property
     def dataset(self) -> xr.Dataset:
--- a/imod/mf6/pkgbase.py
+++ b/imod/mf6/pkgbase.py
@@ -9,6 +9,7 @@ import xugrid as xu
 
 import imod
 from imod.mf6.interfaces.ipackagebase import IPackageBase
+from imod.typing.grid import GridDataset
 
 TRANSPORT_PACKAGES = ("adv", "dsp", "ssm", "mst", "ist", "src")
 EXCHANGE_PACKAGES = "gwfgwf"
@@ -47,11 +48,11 @@ class PackageBase(IPackageBase, abc.ABC):
         return variable_names
 
     @property
-    def dataset(self) -> xr.Dataset:
+    def dataset(self) -> GridDataset:
         return self.__dataset
 
     @dataset.setter
-    def dataset(self, value: xr.Dataset) -> None:
+    def dataset(self, value: GridDataset) -> None:
         self.__dataset = value
 
     def __getitem__(self, key):
--- a/imod/mf6/riv.py
+++ b/imod/mf6/riv.py
@@ -1,3 +1,5 @@
+import inspect
+
 import numpy as np
 
 from imod.mf6.auxiliary_variables import add_periodic_auxiliary_variable
@@ -15,6 +17,8 @@ from imod.schemata import (
     IndexesSchema,
     OtherCoordsSchema,
 )
+
+ARGS_TO_EXCLUDE = ["validate"]
 
 
 class River(BoundaryCondition):
@@ -144,6 +148,14 @@ class River(BoundaryCondition):
         validate: bool = True,
         repeat_stress=None,
     ):
+        locals_at_init = locals()
+        variable_names = list(inspect.signature(self.__init__).parameters.keys())
+        for var_to_exclude in ARGS_TO_EXCLUDE:
+            if var_to_exclude in variable_names:
+                variable_names.remove(var_to_exclude)
+        variables_to_merge = {name: locals_at_init[name] for name in variable_names}
+        xr.merge([variables_to_merge], join="exact")
+
         super().__init__(locals())
         self.dataset["stage"] = stage
         self.dataset["conductance"] = conductance
--- a/imod/mf6/riv.py
+++ b/imod/mf6/riv.py
@@ -146,22 +146,15 @@ class River(BoundaryCondition):
     ):
         locals_at_init = locals()
         variable_names = self._get_variable_names(self.__init__)
+        if concentration is None:
+            variable_names.remove("concentration")
+            variable_names.remove("concentration_boundary_type")
         variables_to_merge = {name: locals_at_init[name] for name in variable_names}
-        xr.merge([variables_to_merge], join="exact")
 
-        super().__init__(locals())
-        self.dataset["stage"] = stage
-        self.dataset["conductance"] = conductance
-        self.dataset["bottom_elevation"] = bottom_elevation
+        super().__init__(variables_to_merge)
         if concentration is not None:
-            self.dataset["concentration"] = concentration
-            self.dataset["concentration_boundary_type"] = concentration_boundary_type
             add_periodic_auxiliary_variable(self)
-        self.dataset["print_input"] = print_input
-        self.dataset["print_flows"] = print_flows
-        self.dataset["save_flows"] = save_flows
-        self.dataset["observations"] = observations
-        self.dataset["repeat_stress"] = repeat_stress
+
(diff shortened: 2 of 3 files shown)


imod/mf6/pkgbase.py
```diff
@@ -8,9 +9,42 @@
 
 import imod
 from imod.mf6.interfaces.ipackagebase import IPackageBase
+from imod.typing.grid import GridDataset
 
 TRANSPORT_PACKAGES = ("adv", "dsp", "ssm", "mst", "ist", "src")
 EXCHANGE_PACKAGES = "gwfgwf"
+ARGS_TO_EXCLUDE = ["validate"]
+
+
```
luitjansl, line 18: would it make sense to relocate these functions out of packagebase and into a separate source file (for example of xugrid utilities)? Or into xugrid? They could be useful in other contexts.

imod/mf6/riv.py
```diff
@@ -144,19 +144,17 @@ def __init__(
         validate: bool = True,
         repeat_stress=None,
     ):
-        super().__init__(locals())
-        self.dataset["stage"] = stage
-        self.dataset["conductance"] = conductance
-        self.dataset["bottom_elevation"] = bottom_elevation
+        locals_at_init = locals()
```
luitjansl, line 147: I assume that with this new functionality, many package constructors will contain the same code (a few lines to merge all the variables and to validate).
So they could be moved to PkgBase, but we need them in the package source file not for the code but for the argument list in the function header.  That is a bit counterintuitive. Another option would be to fill a dictionary in each package source file with the variable names and types its constructor needs. Then the constructor could go into pkgbase, and use these package-specific variables (a few packages would still have a custom contructor but that's fine) 
Or we could make a separate file with details of the datamodel, which lists for each package what goes into the package along with types, validation hints, and preferred regridding method. (In this approach we could also split what validation requirements and function arguments come from Modflow and what requirements and function arguments are imod-python specific). This would clean up the source files considerably, because the package metadata often is a considerable section of the package source file. 

Overall, I like the way constructors can be made shorter with this approach ( the one for the lake package is 60 lines, and with these changes, maybe 6?) but the downside is that it becomes (slightly) less transparent how things work for new developers.

luitjansl (commented):
interesting! I see advantages to the approach. As put out in my comments, I would hope that it can be combined with a reduction of metadata in the source files.

imod/mf6/riv.py
Manangka: A similar approach could be:

```
def __init__(...):
  arg_dict = args2dict(locals())
    if concentration is None:
      del arg_dict["concentration"]
      del arg_dict["concentration_boundary_type"]

      super().__init__(arg_dict)
   ....
```

```
def args2dict(arguments):
    arguments.pop("self", None)
    arguments.pop("__class__", None)

    return arguments

```

A different approach could be to define a decorator to wrap the init method

```
@pkg_init
def __init__(...):
  ...
```


```
def pkg_init(init):
    @wraps(init)
    def merge_init(self, *args, **kwargs):
        args_dict = inspect.signature(self.__init__).bind_partial(*args, **kwargs).arguments
        for k, v in inspect.signature(self.__init__).parameters.items():
            if k not in args_dict and v.default is not inspect.Parameter.empty:
                args_dict[k] = v.default

        super(type(self), self).__init__(args_dict)
        return init(self, **args_dict)

    return merge_init
```

imod/mf6/pkgbase.py
```diff
@@ -8,9 +9,42 @@
...
+        dataset[name] = variable
+
+    return dataset
+
+
+def merge_at_init(variables_to_merge):
+    if variables_to_merge is not None:
+        for arg in variables_to_merge.values():
+            if isinstance(arg, xu.UgridDataArray):
+                return merge_unstructured_dataset(variables_to_merge)
```
Manangka, line 45: This check can be made a bit stricter. We don;t want to mix dataarrays and ugriddataarays

```
def merge_at_init(variables_to_merge):
    if variables_to_merge is None:
        return xr.Dataset()

    if all(isinstance(variable, xr.Dataarray) for variable in variables_to_merge):
        return xr.merge([variables_to_merge], join="exact")
    
    if all(isinstance(variable, xu.UgridDataArray) for variable in variables_to_merge):
        return merge_unstructured_dataset(variables_to_merge)
    
    raise ArgumentError("Mixed GridDataArray types. Unable to merge")
```

imod/mf6/pkgbase.py
```diff
@@ -8,9 +9,42 @@
...
 import imod
 from imod.mf6.interfaces.ipackagebase import IPackageBase
+from imod.typing.grid import GridDataset
 
 TRANSPORT_PACKAGES = ("adv", "dsp", "ssm", "mst", "ist", "src")
 EXCHANGE_PACKAGES = "gwfgwf"
+ARGS_TO_EXCLUDE = ["validate"]
+
+
+def merge_unstructured_dataset(variables_to_merge):
```
Manangka, line 19: I would move this to a separate class. You could add this to grid.py class in the typing folder. And have a typeddispatch for the xr and xu versions of merge. It will also make it easier to remove this functionality when xugrid supports merging

imod/mf6/pkgbase.py
```diff
@@ -8,9 +9,42 @@
 
 import imod
 from imod.mf6.interfaces.ipackagebase import IPackageBase
+from imod.typing.grid import GridDataset
 
 TRANSPORT_PACKAGES = ("adv", "dsp", "ssm", "mst", "ist", "src")
 EXCHANGE_PACKAGES = "gwfgwf"
+ARGS_TO_EXCLUDE = ["validate"]
```
Manangka, line 16: I'm not sure this should be here. It might as well be in the derived packages. You could also add an argument to the _get_variable_names method called args_to_exclude

Manangka (changes requested):

imod/mf6/pkgbase.py
JoerivanEngelen: Removed!

imod/mf6/pkgbase.py
JoerivanEngelen: I now implemented it as a argument to the wrapper

imod/mf6/riv.py
JoerivanEngelen: Thanks for the wrapper idea, I implemented that.

JoerivanEngelen (commented):
Thanks! I for now implemented the wrapper idea, I think it nicely cleans up the code. My only slight doubt is the disadvantage which my initial implemention also had: it appears as arguments / keyword arguments are unused.

JoerivanEngelen:
Just as an extra check to see if xr.merge causes any loss of performance in this case as reported in: https://github.com/Deltares/imod-python/issues/736#issuecomment-1884643780 . Apparently with nicely overlapping coordinates, merging is not such a slow process and the changes made in this PR should only (very slightly) enhance performance.

```python
# %% import
import numpy as np
import xarray as xr


# %% functions
def make_da(nlay, nrow, ncol):
    shape = nlay, nrow, ncol

    dx = 10.0
    dy = -10.0
    xmin = 0.0
    xmax = dx * ncol
    ymin = 0.0
    ymax = abs(dy) * nrow
    dims = ("layer", "y", "x")

    layer = np.arange(1, nlay + 1)
    y = np.arange(ymax, ymin, dy) + 0.5 * dy
    x = np.arange(xmin, xmax, dx) + 0.5 * dx
    coords = {"layer": layer, "y": y, "x": x}

    return xr.DataArray(
        data=np.ones(shape, dtype=float),
        dims=dims,
        coords=coords,
    )


def make_big_data_dict(nvar=4, nlay=3, nrow=1000, ncol=1000):
    varnames = [f"var{i}" for i in np.arange(nvar)]
    return {var: make_da(nlay, nrow, ncol) for var in varnames}


def ds_from_merge_exact(d):
    return xr.merge([d], join="exact")


def ds_from_merge_default(d):
    return xr.merge([d])


def ds_assigned(d):
    ds = xr.Dataset()
    for key, value in d.items():
        ds[key] = value

    return ds


# %% Create data dict

d = make_big_data_dict()

# %% Benchmark
# Benchmark in iPython notebook

# %timeit ds_exact = ds_from_merge_exact(d)
## > 424 µs ± 3.53 µs per loop (mean ± std. dev. of 7 runs, 1,000 loops each)

# %timeit ds_default = ds_from_merge_default(d)
## > 864 µs ± 11.4 µs per loop (mean ± std. dev. of 7 runs, 1,000 loops each)

# %timeit ds = ds_assigned(d)
## > 1.75 ms ± 42.5 µs per loop (mean ± std. dev. of 7 runs, 1,000 loops each)
```

imod/mf6/riv.py
```diff
@@ -144,19 +145,6 @@ def __init__(
         validate: bool = True,
         repeat_stress=None,
     ):
```
Manangka, line 147: Nice and clean!

Manangka (changes requested):

Manangka (approved):

imod/mf6/pkgbase.py
```diff
@@ -1,16 +1,51 @@
...
+    # https://stackoverflow.com/a/42581103
+    def pkg_init_decorator(init: Callable):
+        @wraps(init)
+        def merge_init(self, *args, **kwargs):
+            # Collect user specified args and kwargs
+            args_dict = (
+                inspect.signature(self.__init__).bind_partial(*args, **kwargs).arguments
+            )
+            # Add default values
+            for k, v in inspect.signature(self.__init__).parameters.items():
```
luitjansl, line 32: you can improve clarity in this function 

- by adding comments. What are you looping over ( the arguments of the constructor being wrapped?)

- k, v = key , value? if so, write it out for clarity. What are their types? (since below we call v.default)

imod/mf6/pkgbase.py
```diff
@@ -1,16 +1,51 @@
...
 import imod
 from imod.mf6.interfaces.ipackagebase import IPackageBase
+from imod.typing.grid import GridDataset, merge_with_dictionary
 
 TRANSPORT_PACKAGES = ("adv", "dsp", "ssm", "mst", "ist", "src")
 EXCHANGE_PACKAGES = "gwfgwf"
+ARGS_TO_EXCLUDE = ["validate"]
+
+
+def pkg_init(exclude_in_dataset: list[str]):
```
luitjansl, line 21: add the return types for these functions.

imod/mf6/pkgbase.py
```diff
@@ -25,20 +60,17 @@ class PackageBase(IPackageBase, abc.ABC):
...
         return super(PackageBase, cls).__new__(cls)
 
-    def __init__(self, allargs=None):
-        if allargs is not None:
-            for arg in allargs.values():
-                if isinstance(arg, xu.UgridDataArray):
-                    self.__dataset = xu.UgridDataset(grids=arg.ugrid.grid)
-                    return
-        self.__dataset = xr.Dataset()
+    def __init__(self, variables_to_merge=None):
```
luitjansl, line 63: add type hints for variables_to_merge

imod/mf6/boundary_condition.py
```diff
@@ -64,6 +67,15 @@ class BoundaryCondition(Package, abc.ABC):
     not the array input which is used in :class:`Package`.
     """
 
+    def __init__(self, allargs=None):
+        super().__init__(allargs)
+        if allargs["concentration"] is None:
```
luitjansl, line 72: this is a strange operation. please add a comment as to why it is necessary ( how did "concentration" end up in self.dataset if it was not in the constructor arguments?). Did the wrapper put it in there? But the wrapper used is a property of the derived class, not of boundary_condition.

imod/mf6/riv.py
```diff
@@ -130,6 +130,7 @@ class River(BoundaryCondition):
         "concentration": (RegridderType.OVERLAP, "mean"),
     }
 
+    @pkg_init(exclude_in_dataset=["validate"])
```
luitjansl, line 133: this makes me wonder how the wrapping will work if we create derived types of the river class. Will we automatically get the same wrapper when we call super(*).__init__ ? can we remove the wrapper if we don't want it? if the new class also uses the @pkg_init wrapper, will it be done twice?

luitjansl (approved):
looks good, some questions and comments

imod/mf6/pkgbase.py
```diff
@@ -1,16 +1,51 @@
...
 import xarray as xr
 import xugrid as xu
 
 import imod
 from imod.mf6.interfaces.ipackagebase import IPackageBase
+from imod.typing.grid import GridDataset, merge_with_dictionary
 
 TRANSPORT_PACKAGES = ("adv", "dsp", "ssm", "mst", "ist", "src")
 EXCHANGE_PACKAGES = "gwfgwf"
+ARGS_TO_EXCLUDE = ["validate"]
```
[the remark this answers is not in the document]
JoerivanEngelen: Woops, good catch!

JoerivanEngelen:
UPDATE:

Though the ``pkg_init`` decorator made everything look nice and clean, I reverted using it as it had three downsides:

* Upon failure the decorator threw cryptic errors messages
* It appears as if arguments are unused (also causing linting errors)
* It complicates our codebase more, potentially scaring off potential new developers

Next to that, I had stackoverflow errors for certain tests, but those probably could be fixed with some work.

For now, I think we are better off forwarding arguments in dictionaries.

Manangka (approved):

imod/mf6/boundary_condition.py
```diff
@@ -64,6 +67,15 @@ class BoundaryCondition(Package, abc.ABC):
     not the array input which is used in :class:`Package`.
     """
 
+    def __init__(self, allargs=None):
+        super().__init__(allargs)
+        if "concentration" in allargs.keys() and allargs["concentration"] is None:
+            # Remove vars inplace
+            del self.dataset["concentration"]
+            del self.dataset["concentration_boundary_type"]
+        else:
```
luitjansl, line 76: if concentration is not present in allargs.keys(), should we do the expand_transient_auxiliary_variables?

luitjansl (approved):
looks good

imod/mf6/boundary_condition.py
luitjansl: but now expand_transient_auxiliary_variables(self) is in the else branch. It will do it if concentration is not present in allargs.keys()

Issue #622, opened 2023-11-03
Cleanup utilities part 1


Manangka:
In GitLab by @JoerivanEngelen on Nov 3, 2023, 11:17

Currently we have an ``imod.util``, ``imod.wq.timeutil``, ``imod.mf6.utilities``. 

There are multiple issues with the current state:

- It is a mix of utilities which are used internally and which are useful for users. 
- ``imod.util`` contains utilities for handling time, space, and paths. I think we can split this up into separate utilities.
- ``imod.wq.timeutil`` contains useful utilities which are also used in the ``mf6`` and ``flow`` module. So these can be moved one level up.

The first step is: to create topical utilities: ``imod.util.grid``, ``imod.util.paths``, ``imod.util.time``. In the ``util/__init__.py`` we can expose utilities which are useful for users as well (and maintain backwards compatibility).

JoerivanEngelen:
Related to https://github.com/Deltares/imod-python/issues/542

PR #723, opened 2024-01-05
Issue #622 properly compute aux vars cdist and anglex for unstructured grids

luitjansl:
This PR introduces tests for the geometric coefficient angledegx which is computed at partition boundaries. Angledegx is used for specific discharge calculations, hence the testing for specific discharge results before and after splitting a model. An example was also added showing how both flopy and imod-python can be used to do the model splitting.


imod/mf6/multimodel/exchange_creator_unstructured.py
```diff
@@ -70,8 +83,30 @@ def _compute_geometric_information(self) -> pd.DataFrame:
...
+        inprod = (
+            normal[:, 0] * outward_vector[:, 0] + normal[:, 1] * outward_vector[:, 1]
+        )
+
+        normal = np.where(
+            np.transpose((inprod[:] > 0, inprod[:] > 0)), normal[:], -normal[:]
+        )
+
+        # using the normal and the norm of the normal we can compute the angle of the normal
+        norm_normal = np.sqrt(normal[:, 0] * normal[:, 0] + normal[:, 1] * normal[:, 1])
```
Huite, line 104: You can use linalg.norm, just like in line 84

imod/mf6/multimodel/exchange_creator_unstructured.py
```diff
@@ -57,10 +57,23 @@ def _compute_geometric_information(self) -> pd.DataFrame:
...
-        face1, face2 = edge_face_connectivity[self._connected_cell_edge_indices].T
+        unordered_face1, unordered_face2 = edge_face_connectivity[
+            self._connected_cell_edge_indices
+        ].T
+        face_partition_1 = self._get_partition_numbers(unordered_face1)
+        face_partition_2 = self._get_partition_numbers(unordered_face2)
+        face1 = np.where(
+            face_partition_1 > face_partition_2, unordered_face2, unordered_face1
+        )
+        face2 = np.where(
```
Huite, line 68: I had to do some thinking about why the order of the faces matters, i.e. why does it matter what is face1 and what is face2? After reading the code hereafter and thinking about it, I guess it's because in the exchanges there is a first and second partition too? And in that case, you'd want the orientation to point in the right direction?

At any rate, it's not obvious to me, so it would be nice to write some comments here.

imod/mf6/multimodel/exchange_creator_unstructured.py
```diff
@@ -70,8 +83,30 @@ def _compute_geometric_information(self) -> pd.DataFrame:
...
+        # vertex index, coordinate index, point 1 or point 2
+        dx = edge_coordinates[:, 1, 0] - edge_coordinates[:, 0, 0]
+        dy = edge_coordinates[:, 1, 1] - edge_coordinates[:, 0, 1]
+
+        normal = np.array((dy[:], -dx[:]), dtype=np.float_).T
+        outward_vector = centroid_2 - centroid_1
+
+        # the inproduct between the edge normal and the outward vector is positive if they go in the same direction.
+        # we use this to swap the normal if needed so that it goes outward.
+        inprod = (
```
Huite, line 95: `np.dot` would also do the trick?

JoerivanEngelen (commented):
Thanks for the involved testing and fixing this! 

Apart from the comments made by Huite, I don't see anything that needs changing.

I like the addition of all the cases for the specific discharge. I think we should merge these cases with the cases already made in the other module with unstructured tests for multimodel.

imod/mf6/multimodel/exchange_creator_unstructured.py
```diff
@@ -70,8 +83,30 @@ def _compute_geometric_information(self) -> pd.DataFrame:
...
+            normal[:, 0] * outward_vector[:, 0] + normal[:, 1] * outward_vector[:, 1]
+        )
```
Manangka, line 97: I'm a bit surprised this is needed. I would expect that something like this should be sufficient:
```
# Normal vector w.r.t. to face1
normal = np.cross(U, [0, 0, -1])[:, :-1]

# Faces that are flipped
flipped_faces = self._get_partition_numbers(face1) > self._get_partition_numbers(face2)

# Correct normal for the partition numbering
normal[flipped_faces] = -normal[flipped_faces]
```

Could it be that we are not following the vertex defintion as described in the modflow manual i.e. that the elements are described by a list of point in a clockwise way. Because if that is the case then the cross product should always return the outward/normal vector. For some reason that is not happening now

Manangka (changes requested):

imod/mf6/multimodel/exchange_creator_unstructured.py
luitjansl: done

imod/mf6/multimodel/exchange_creator_unstructured.py
luitjansl: it is indeed as you say: we want the exchange from the partition with the lower number to that with the higher number. I added a comment to that effect.

imod/mf6/multimodel/exchange_creator_unstructured.py
```diff
@@ -70,8 +83,30 @@ def _compute_geometric_information(self) -> pd.DataFrame:
         Vj = centroid_2 - edge_coordinates[:, 0]
```
Manangka, line 83: By changing the face1 and face2 on lines 65 and 68 you might flip around centroid_1 and centroid_2 which in turn have affect on the computed Vi and Vj and thus the cl1 and cl2 values. By flipping those values cl1 and cl2 might be flipped incorrectly

imod/mf6/multimodel/exchange_creator_unstructured.py
```diff
@@ -70,8 +83,30 @@ def _compute_geometric_information(self) -> pd.DataFrame:
...
         length = np.linalg.norm(U, axis=1)
 
-        # The cross product of U and V equals the area of the parallellogram,
-        # dividing by length (the base of the parallellogram) gives the orthogonal distances cl1 & cl2.
+        # vertex index, coordinate index, point 1 or point 2
+        dx = edge_coordinates[:, 1, 0] - edge_coordinates[:, 0, 0]
```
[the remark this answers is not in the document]
luitjansl: indeed; now I'm using this.

imod/mf6/multimodel/exchange_creator_unstructured.py
luitjansl: that does not seem to work: for 2d arrays, np.dot does a matrix multiplication

imod/mf6/multimodel/exchange_creator_unstructured.py
```diff
@@ -57,10 +57,23 @@ def _compute_geometric_information(self) -> pd.DataFrame:
         grid = self._submodel_labels.ugrid.grid
         edge_face_connectivity = grid.edge_face_connectivity
 
-        face1, face2 = edge_face_connectivity[self._connected_cell_edge_indices].T
+        unordered_face1, unordered_face2 = edge_face_connectivity[
```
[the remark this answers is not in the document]
luitjansl: done

imod/mf6/multimodel/exchange_creator_unstructured.py
```diff
@@ -70,8 +83,30 @@ def _compute_geometric_information(self) -> pd.DataFrame:
...
+        )
+
+        normal = np.where(
+            np.transpose((inprod[:] > 0, inprod[:] > 0)), normal[:], -normal[:]
+        )
+
+        # using the normal and the norm of the normal we can compute the angle of the normal
+        norm_normal = np.sqrt(normal[:, 0] * normal[:, 0] + normal[:, 1] * normal[:, 1])
+        cos = (normal[:, 0]) / norm_normal
+        angle = np.degrees(np.arccos(cos))
```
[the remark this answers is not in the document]
luitjansl: good point, done

imod/mf6/multimodel/exchange_creator_unstructured.py
Manangka: I did some more digging and found out why the code snippet above doesn't work. It has to do with the fact that the edge-node_coordinates don't follow the same convention and used to define an element. In xugrid the vertex order of an element is defined to be counterclockwise, but the edge-node_coordinates doesn't take this into account.

I wrote the following snippet that checks of the edge coordinates follow the same direction as the element vertex order and if not it flips them around. This allows us to simplfy this code and make it easier to understand.

@Huite Can we enforce a correct ordering of the edge-node_coordinates in xugrid?

 ```
      def _compute_geometric_information(self) -> pd.DataFrame:
        grid = self._submodel_labels.ugrid.grid
        edge_face_connectivity = grid.edge_face_connectivity

        face1, face2 = edge_face_connectivity[self._connected_cell_edge_indices].T

        flipped_faces_mask = self._get_partition_numbers(face1) > self._get_partition_numbers(face2)
        centroid_1 = grid.centroids[face1]
        centroid_2 = grid.centroids[face2]

        # I don't know why this is necessary
        temp = centroid_1[flipped_faces_mask]
        centroid_1[flipped_faces_mask] = centroid_2[flipped_faces_mask]
        centroid_2[flipped_faces_mask] = temp

        cdist = np.linalg.norm(centroid_2 - centroid_1, axis=1)

        edge_coordinates = grid.edge_node_coordinates[self._connected_cell_edge_indices]

        U = np.diff(edge_coordinates, axis=1)[:, 0]
        # Compute vector of first cell centroid to first edge vertex
        Vi = centroid_1 - edge_coordinates[:, 0]
        # Compute vector of second cell centroid to first edge vertex
        Vj = centroid_2 - edge_coordinates[:, 0]
        length = np.linalg.norm(U, axis=1)

        clockwise_line_mask = [self.is_clockwise_line(face, edge_coordinates[idx, :]) for idx, face in enumerate(face1)]

        normal = np.cross(U, [0, 0, -1])[:, :-1]
        normal[clockwise_line_mask] = -normal[clockwise_line_mask]
        normal[flipped_faces_mask] = -normal[flipped_faces_mask]
        angle = np.degrees(np.arctan2(normal[:, 1], normal[:, 0]))

        df = pd.DataFrame(
            {
                "cell_idx1": self._connected_cells["cell_idx1"].values,
                "cell_idx2": self._connected_cells["cell_idx2"].values,
                "cl1": np.abs(np.cross(U, Vi)) / length,
                "cl2": np.abs(np.cross(U, Vj)) / length,
                "hwva": length,
                "angldegx": angle,
                "cdist": cdist,
            }
        )
        return df

    def is_clockwise_line(self, faceid, edge):
        # The elements in xu grid are defined counterclockwise. However the line provided
        # in the grid.edge_node_coordinates don't follow the same convention. This method verifies if the edge follow
        # the same directions as the elements definition

        from shapely.geometry import Polygon, LineString
        from shapely.ops import shared_paths

        x = self._submodel_labels.ugrid.grid.face_node_coordinates[faceid][:, 0]
        y = self._submodel_labels.ugrid.grid.face_node_coordinates[faceid][:, 1]

        poly = Polygon(zip(x, y))
        line = LineString(zip(edge[:, 0], edge[:, 1]))

        forward, backward = shared_paths(poly.exterior, line).geoms
        if not forward.is_empty:
            return True
        if not backward.is_empty:
            return False

        raise ValueError("Disjoint line")
```

**Review by Manangka — CHANGES_REQUESTED** [review]

File: imod/mf6/multimodel/exchange_creator_unstructured.py (hunk 5 above)
Comment (luitjansl) [reply imod/mf6/multimodel/exchange_creator_unstructured.py:83]: indeed, they can be flipped. But in a good way: cl1 refers to the distance from the cell centre of the cell  with the lower partition number to the shared edge; cl2 to that of the cell centre with the higher partition number to that edge. Now it is computed like that.

File: imod/mf6/multimodel/exchange_creator_unstructured.py (hunk 4 above)
Comment (Huite) [reply imod/mf6/multimodel/exchange_creator_unstructured.py:84]: @Manangka We could open an issue on the xugrid repo for this.

I don't think xugrid can enforce the orientation. I just check the UGRID conventions page and looked for "clockwise":
https://ugrid-conventions.github.io/ugrid-conventions/

It only mentions that the face_node_connectivity should be counterclockwise, as well as the face_edge_connectivity.
The edge_node_connectivity is used for 1D topologies, in which case the direction is meaningful if we interpret it a a directed graph (e.g. a flow network). That might be reason why they didn't prescribe anything in the conventions.

Anyway, the edge_node_connectivity definition is fully up to the user and they are currently free to attach meaning to it, which means I wouldn't like xugrid to change it. I also haven't come across other uses for the orientation of the edges yet, so I concur with @luitjansl.

On a different note, shapely is quite slow if you're dealing with simple convex geometries in my experience.

Comment (Manangka) [reply imod/mf6/multimodel/exchange_creator_unstructured.py:84]: I see. So we have to find the direction of the normal ourself.

If we take the inner product of the normal vector with a vector from the edge to the face centroid then the normal faces inwards if the value is greater than 0. 

   ```
 def _compute_geometric_information(self) -> pd.DataFrame:
        grid = self._submodel_labels.ugrid.grid
        face1, face2 = self._get_partition_sorted_connected_faces()
        centroid_1 = grid.centroids[face1]
        centroid_2 = grid.centroids[face2]

        cdist = np.linalg.norm(centroid_2 - centroid_1, axis=1)

        edge_coordinates = grid.edge_node_coordinates[self._connected_cell_edge_indices]

        U = np.diff(edge_coordinates, axis=1)[:, 0]
        # Compute vector of first cell centroid to first edge vertex
        Vi = centroid_1 - edge_coordinates[:, 0]
        # Compute vector of second cell centroid to first edge vertex
        Vj = centroid_2 - edge_coordinates[:, 0]
        length = np.linalg.norm(U, axis=1)

        normal = np.cross(U, [0, 0, -1])[:, :-1]
        # If the inner product of the normal with a vector on the edge to the face centroid is positive
        # then the normal vector points inwards
        inward_vector_mask = np.sum(normal * Vi, axis=-1) > 0
        normal[inward_vector_mask] = -normal[inward_vector_mask]

        angle = np.degrees(np.arctan2(normal[:, 1], normal[:, 0]))

        df = pd.DataFrame(
            {
                "cell_idx1": self._connected_cells["cell_idx1"].values,
                "cell_idx2": self._connected_cells["cell_idx2"].values,
                "cl1": np.abs(np.cross(U, Vi)) / length,
                "cl2": np.abs(np.cross(U, Vj)) / length,
                "hwva": length,
                "angldegx": angle,
                "cdist": cdist,
            }
        )
        return df
```

luitjansl: done, now using Vi to determine the direction of the normal. I kept the normal calculation based on the components of the U vector.

Manangka (approved):
Looks good!

Issue #534, opened 2023-09-07
Add test to test if HFB is not written to binary


Manangka:
In GitLab by @JoerivanEngelen on Sep 7, 2023, 11:33

As it stands, Modflow 6 doesn't support binary files for the HFB package.
Would be good to add a test to test if this doesn't happen accidently, 
such as [this as issue](https://gitlab.com/deltares/imod/imod-python/-/issues/523)

PR #725, opened 2024-01-05
Issue #534 Test if hfb not written to binary

JoerivanEngelen:
Close https://github.com/Deltares/imod-python/issues/534

I noticed that the ``test_hfb_writing_one_layer`` would fail if the HFB is written accidently to binary. This adds explicit assertions if not written to binary files. In the case a binary file is written as ``.dat`` file, still an error is thrown with in ``np.loadtxt``, such as:

```
ValueError: could not convert string '\x01\x00\x00\x00\x01\x00\x00\x00\x01\x00\x00\x00\x03\x00\x00\x00ü©ñÒMbP?\x01\x00\x00\x00\x02\x00\x0 to float64 at row 0, column 1.
```

This should point us quickly enough to the source of the error.

### DIFF
--- a/imod/tests/test_mf6/test_mf6_hfb_adapter.py
+++ b/imod/tests/test_mf6/test_mf6_hfb_adapter.py
@@ -118,5 +118,13 @@ def test_hfb_writing_one_layer(barrier, tmp_path):
     hfb.write("hfb", None, write_context)
 
     # Assert
-    data = np.loadtxt(tmp_path / "hfb" / "hfb.dat")
+    hfb_bin_path = tmp_path / "hfb" / "hfb.bin"
+    hfb_txt_path = tmp_path / "hfb" / "hfb.dat"
+
+    # Test if not written to binary file
+    assert not hfb_bin_path.exists()
+    # Test if text file exists
+    assert hfb_txt_path.exists()
+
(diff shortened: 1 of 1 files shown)


imod/tests/test_mf6/test_mf6_hfb_adapter.py
```diff
@@ -118,5 +118,13 @@ def test_hfb_writing_one_layer__unstructured(barrier, tmp_path):
     hfb.write("hfb", None, write_context)
 
     # Assert
-    data = np.loadtxt(tmp_path / "hfb" / "hfb.dat")
+    hfb_bin_path = tmp_path / "hfb" / "hfb.bin"
+    hfb_txt_path = tmp_path / "hfb" / "hfb.dat"
+
+    # Test if not written to binary file
```
Manangka, line 124: I would remove this comment and the one on line 126. I think it is clear from the code what it does.
If you do want to add that info you could consider renaming the test name to something like `test_hfb_writing_one_layer_non_binary_format`

Manangka (approved):

imod/tests/test_mf6/test_mf6_hfb_adapter.py
JoerivanEngelen: Done

PR #733, opened 2024-01-09
Issue #732 Fix CodeCoverage chart which point at wrong buildType.

Manangka:
A buildchart on the project page which is responsible for displaying the codecoverage results was pointing at the wrong buildType. This commit fix that.

Further more there is also a buildReportTab for displaying the codecoverage on a build level. However this is not necessary. When a coverage.zip artifact is found in an build this happens autimatically


JoerivanEngelen (approved):
Looks good to me!

Issue #682, opened 2023-12-11
Fix Mypy issues in the model files


Manangka:
In GitLab by @Manangka on Dec 7, 2023, 1:37

Issue #626, opened 2023-11-07
Clip_box_packages() passes wrong arguments to wel package clip_box()


Manangka:
In GitLab by @HendrikKok on Nov 7, 2023, 10:27

_clip_box_packages() in model.py passes ‘layer_min’ and ‘layer_max’ as arguments. The package clip_box of well-class expects ‘z_min’ and ‘z_max’.

Pragmatic solution:

* Special-case gridagnostic packages (HFB non layered, WEL)
* Find z_min and z_max based on layer_min and layer_max in DIS package
* Provide that to clip_box
 
This requires the following refactoring/addition:

* The grid agnostic packages should accept layer_min and layer_max scalar values. ``cellids`` are determined in the function and consequently filtered whether they exceed layer_min and layer_max.

### Tasks
- [x] https://github.com/Deltares/imod-python/issues/681
- [x] https://github.com/Deltares/imod-python/issues/682
- [x] https://github.com/Deltares/imod-python/issues/684
- [ ] #740

Manangka:
added #682 as child task

Manangka:
added #684 as child task

PR #739, opened 2024-01-10
Issue #626 Fix MyPy errors

### LINKED ISSUE #681: Move the GroundWaterFlow a GrounWaterTransport models to separate files
In GitLab by @Manangka on Dec 7, 2023, 1:36

Manangka:
This also fixes #681, #682, #684


imod/mf6/package.py
```diff
@@ -40,8 +40,9 @@ class Package(PackageBase, abc.ABC):
     """
 
     _pkg_id = ""
-    _init_schemata = {}
```
JoerivanEngelen, line 43: I don't think ``ValueSchema`` is the right type here, as that is quite a specific implementation. I think we need to define a ``SchemaType`` or something for this.

JoerivanEngelen (changes requested):
This cleans up things nicely and better type annotation is good. I have one comment on the type annotation used for schemata.

imod/mf6/model.py
```diff
@@ -501,6 +531,10 @@ def _get_regridding_domain(
                 included_in_all = regridded_idomain
             else:
                 included_in_all = included_in_all.where(regridded_idomain.notnull())
+
+        if included_in_all is None:
+            raise ValueError("Unable to regrid")
```
luitjansl, line 536: pleas expand the error message a bit to explain why is is unable to regrid

imod/util.py
```diff
@@ -1159,7 +1159,7 @@ def _replace(
     )
 
 
-def values_within_range(da, min=None, max=None):
+def values_within_range(da, min: Optional[float] = None, max: Optional[float] = None):
```
luitjansl, line 1162: return type missing

luitjansl (approved):
looks good

imod/mf6/package.py
JoerivanEngelen: I means defining something like this in ``schemata.py``

```python
from typing import TypeVar

SchemaType = TypeVar("SchemaType", bound="BaseSchema")
```

Manangka: Applied this

JoerivanEngelen (approved):
Looks good! Bit of a shocker that we forgot to add ``fastcore`` to the ``pyproject.toml`` :-O

PR #741, opened 2024-01-10
Issue #740 Fix incorrect arguments provided to the clip_box method of the well package

Manangka:
In this commit the wrong arguments provided to the well package is addressed. 

The Well package is a grid agnostic package. That means that the layer_min and layer_max arguments from the clip_box method has no meaning to it. 

To still conform to this method signature the top and bottom of a grid is provided. Those are then used to retrieve a z_min and z_max

### DIFF
--- a/imod/mf6/package.py
+++ b/imod/mf6/package.py
@@ -1,3 +1,5 @@
+from __future__ import annotations
+
 import abc
 import copy
 import numbers
@@ -401,16 +403,18 @@ class Package(PackageBase, abc.ABC):
 
     def clip_box(
         self,
-        time_min=None,
-        time_max=None,
-        layer_min=None,
-        layer_max=None,
-        x_min=None,
-        x_max=None,
-        y_min=None,
-        y_max=None,
-        state_for_boundary=None,
-    ) -> "Package":
+        time_min: Optional[str] = None,
+        time_max: Optional[str] = None,
+        layer_min: Optional[int] = None,
+        layer_max: Optional[int] = None,
+        x_min: Optional[float] = None,
+        x_max: Optional[float] = None,
+        y_min: Optional[float] = None,
+        y_max: Optional[float] = None,
+        top: Optional[GridDataArray] = None,
+        bottom: Optional[GridDataArray] = None,
+        state_for_boundary: Optional[GridDataArray] = None,
+    ) -> Package:
         """
         Clip a package by a bounding box (time, layer, y, x).
 
@@ -433,6 +437,10 @@ class Package(PackageBase, abc.ABC):
         x_max: optional, float
         y_min: optional, float
         y_max: optional, float
+        top: optional, GridDataArray
+        bottom: optional, GridDataArray
+        state_for_boundary: optional, GridDataArray
+
 
         Returns
         -------
--- a/imod/mf6/package.py
+++ b/imod/mf6/package.py
@@ -403,8 +403,8 @@ class Package(PackageBase, abc.ABC):
 
     def clip_box(
         self,
-        time_min: Optional[str] = None,
-        time_max: Optional[str] = None,
+        time_min: Optional[cftime.datetime | np.datetime64 | str] = None,
+        time_max: Optional[cftime.datetime | np.datetime64 | str] = None,
         layer_min: Optional[int] = None,
         layer_max: Optional[int] = None,
         x_min: Optional[float] = None,
--- a/imod/mf6/wel.py
+++ b/imod/mf6/wel.py
@@ -246,7 +246,7 @@ class Well(BoundaryCondition, IPointDataPackage):
         sliced : Package
         """
         if (layer_max or layer_min) and (top is None or bottom is None):
-            raise ValueError("Both the layer_max and the top should be defined")
+            raise ValueError("When clipping by layer both the top and bottom should be defined")
 
         if top is not None:
             if not isinstance(top, GridDataArray) or "layer" not in top.coords:
@@ -260,9 +260,16 @@ class Well(BoundaryCondition, IPointDataPackage):
         z_max = None if layer_max is None else top.isel(layer=layer_max)
         z_min = None if layer_min is None else bottom.isel(layer=layer_min)
 
-        if z_max:
+        # if z is a grid select the the values at the well locations and drop the dimensions
+        if (z_max is not None) and ("x" in z_max.coords or "y" in z_max.coords):
+            z_max = z_max.sel(x=ds["x"], y=ds["y"], method="nearest").drop_vars(lambda x: x.coords)
+
+        if (z_min is not None) and ("x" in z_min.coords or "y" in z_min.coords):
+            z_min = z_min.sel(x=ds["x"], y=ds["y"], method="nearest").drop_vars(lambda x: x.coords)
+
+        if z_max is not None:
             ds["screen_top"] = ds["screen_top"].clip(None, z_max)
-        if z_min:
+        if z_min is not None:
             ds["screen_bottom"] = ds["screen_bottom"].clip(z_min, None)
 
         # Initiate array of True with right shape to deal with case no spatial
--- a/imod/mf6/wel.py
+++ b/imod/mf6/wel.py
@@ -260,14 +260,21 @@ class Well(BoundaryCondition, IPointDataPackage):
         z_max = None if layer_max is None else top.isel(layer=layer_max)
         z_min = None if layer_min is None else bottom.isel(layer=layer_min)
 
+        if z_max:
+            ds["screen_top"] = ds["screen_top"].clip(None, z_max)
+        if z_min:
+            ds["screen_bottom"] = ds["screen_bottom"].clip(z_min, None)
+
         # Initiate array of True with right shape to deal with case no spatial
         # selection needs to be done.
         in_bounds = np.full(ds.dims["index"], True)
         # Select all variables along "index" dimension
         in_bounds &= values_within_range(ds["x"], x_min, x_max)
         in_bounds &= values_within_range(ds["y"], y_min, y_max)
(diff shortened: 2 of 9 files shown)


imod/tests/test_mf6/test_mf6_wel.py
```diff
@@ -137,35 +138,51 @@ def test_to_mf6_pkg__high_lvl_transient(basic_dis, well_high_lvl_test_data_trans
     np.testing.assert_equal(mf6_ds["rate"].values, rate_expected)
 
 
-def test_clip_box__high_lvl_stationary(well_high_lvl_test_data_stationary):
+@pytest.mark.parametrize(
```
luitjansl, line 141: it would be nice to add test cases where for example layer_min is not None, but layer_max is. And vice versa. Or is that combination not allowed?
Also the parameters time_min and time_max could be included in the testing

imod/mf6/package.py
```diff
@@ -401,16 +403,18 @@ def __to_datetime(self, time, use_cftime):
...
+        x_min: Optional[float] = None,
+        x_max: Optional[float] = None,
+        y_min: Optional[float] = None,
+        y_max: Optional[float] = None,
+        top: Optional[GridDataArray] = None,
+        bottom: Optional[GridDataArray] = None,
+        state_for_boundary: Optional[GridDataArray] = None,
+    ) -> Package:
         """
         Clip a package by a bounding box (time, layer, y, x).
```
luitjansl, line 419: Maybe the docstring should be expanded a little to explain that if layer_min is present, bottom should also be present, and if layer_max is present, top should be present.

imod/mf6/package.py
```diff
@@ -401,16 +403,18 @@ def __to_datetime(self, time, use_cftime):
...
-        time_max=None,
-        layer_min=None,
-        layer_max=None,
-        x_min=None,
-        x_max=None,
-        y_min=None,
-        y_max=None,
-        state_for_boundary=None,
-    ) -> "Package":
+        time_min: Optional[str] = None,
```
luitjansl, line 406: is str the type of t_min? Isn't there a more precise type that could be used? Maybe here some input checking can also be added to make sure t_min has a string value that makes sense.

luitjansl (commented):
looks good, some comments

imod/mf6/wel.py
```diff
@@ -237,6 +243,9 @@ def clip_box(
 
         ds = new.dataset
 
+        z_max = None if layer_max is None else top.isel(layer=layer_max)
```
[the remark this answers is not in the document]
JoerivanEngelen: Before this changeset, we expected ``z_max`` and ``z_min`` as a scalar. 
This makes it easy to make simple selections in the table of wells.
This was not feature complete, more a pragmatic workaround so that I not had to deal with locating wells in the grid.

However, top and bottom most of the time are grids. ``z_min``/``z_max`` therefore cannot be expected to be scalar anymore, which means well points have to be located in the grid to do the selecting (at least if ``top`` and ``bottom`` are provided as grids). 
I think we can use ``imod.prepare.wells.locate_wells`` for this. With this we can discard wells easily which are outside top and bottom.

Furthermore: ``top`` can also be provided as scalar by the user, I'm not sure if that perpetuates to the ``clip_box`` method here, or that ``top`` has a got a layer dim assigned to it in the mean time.

imod/tests/test_mf6/test_mf6_wel.py
JoerivanEngelen: I see we are now also using testdata where bottom, top are not grids. In the fixtures: 

```python
@pytest.fixture
def parameterizable_basic_dis(request):
...
    top = xr.DataArray(z[:-1], coords={"layer": layers})
    bottom = xr.DataArray(z[1:], coords={"layer": layers})
...
```

This is possible, but doesn't cover most of the use cases. Most of the applications have ``top`` and ``bottom`` specified as grids. We should cover that in our tests as well.

JoerivanEngelen (changes requested):
2 comments, I think we have to carefully look what dimensions ``top`` and ``bottom`` are/should be provided with here. As layers or 3Dgrid ? And in case of ``top`` as scalar or as 2D grid?

imod/mf6/wel.py
```diff
@@ -220,31 +231,50 @@ def clip_box(
...
+        z_min = None if layer_min is None else bottom.isel(layer=layer_min)
+
+        if z_max:
+            ds["screen_top"] = ds["screen_top"].clip(None, z_max)
+        if z_min:
+            ds["screen_bottom"] = ds["screen_bottom"].clip(z_min, None)
```
JoerivanEngelen, line 266: I don't think this call to ``clip`` works in the case that ``top`` and ``bottom`` is a grid, since wells are not defined as a grid.

imod/tests/test_mf6/test_mf6_wel.py
```diff
@@ -137,35 +140,120 @@ def test_to_mf6_pkg__high_lvl_transient(basic_dis, well_high_lvl_test_data_trans
     np.testing.assert_equal(mf6_ds["rate"].values, rate_expected)
 
 
-def test_clip_box__high_lvl_stationary(well_high_lvl_test_data_stationary):
+class ClipBoxCases:
```
JoerivanEngelen, line 143: I like these cases! I think you should add some cases where ``top`` and ``bottom`` are provided as grids to see what I'm talking about in my other comment ;-)

FYI: In the ``parameterizable_basic_dis`` fixture, ``top`` and ``bottom`` are not created as grids:

```python
    top = xr.DataArray(z[:-1], coords={"layer": layers})
    bottom = xr.DataArray(z[1:], coords={"layer": layers})
```

JoerivanEngelen (changes requested):
The changes made are an improvement, but I think you should add some test cases with ``top`` and ``bottom`` as grids to see what I'm talking about ;-)

imod/mf6/wel.py
```diff
@@ -220,31 +231,63 @@ def clip_box(
...
+        if (z_min is not None) and ("x" in z_min.coords or "y" in z_min.coords):
+            z_min = z_min.sel(x=ds["x"], y=ds["y"], method="nearest").drop_vars(
+                lambda x: x.coords
+            )
+
+        if z_max is not None:
+            ds["screen_top"] = ds["screen_top"].clip(None, z_max)
+        if z_min is not None:
+            ds["screen_bottom"] = ds["screen_bottom"].clip(z_min, None)
+
```
JoerivanEngelen, line 280: Already better, but there are still some issues: 

1. This doesn't work for unstructured grids, these do not have "x" and "y" in coords, so then the if statements fails, as well as the call to ``.sel``, this should be ``.ugrid.sel``. Luckily we don't have to reinvent the wheel. We already wrote ``imod.typing.grid.is_spatial_2D`` for this and ``imod.select.points_values``.
2.  ``z_min`` and ``z_max`` can be None, but can also be first a grid and then be converted to a 1d array. I think it is best to keep ``z_min``, ``z_max`` to something that is accepted by ``values_within_range``, so not as grid.

In conclusion: 

```suggestion
        z_max_grid = None if layer_max is None else top.isel(layer=layer_max)
        z_min_grid = None if layer_min is None else bottom.isel(layer=layer_min)

        # if z is a grid select the the values at the well locations and drop the dimensions
        if (z_max_grid is not None) and is_spatial_2D(z_max_grid):
            z_max =  imod.select.points_values(z_max_grid, x=ds["x"].values, y=ds["y"].values, out_of_bounds="ignore")
            )
        else: 
            z_max = None

        if ( z_min_grid is not None) and is_spatial_2D(z_min_grid):
            z_min =  imod.select.points_values(z_min_grid, x=ds["x"].values, y=ds["y"].values, out_of_bounds="ignore")
            )
        else: 
            z_min = None
```

The call to ``is_spatial_2D(z_max_grid)`` is probably overdoing things, so you can probably remove that one.

imod/tests/test_mf6/test_mf6_wel.py
```diff
@@ -137,35 +141,141 @@ def test_to_mf6_pkg__high_lvl_transient(basic_dis, well_high_lvl_test_data_trans
...
+        top = 0.0
+        clip_arguments = {"layer_max": 2, "bottom": bottom, "top": top}
+
+        expected_dims = {"index": 4, "species": 2}
+        return clip_arguments, expected_dims, does_not_raise()
+
+    @staticmethod
+    def case_clip_top_is_non_layered_griddataarray(parameterizable_basic_dis):
+        _, top, bottom = parameterizable_basic_dis
+        top = top.isel(layer=0).drop_vars("layer")
```
JoerivanEngelen, line 192: I don't see top here defined as griddataarray, missing the call to ``broadcast_to_full_domain here``.

JoerivanEngelen (changes requested):
Looks better, I have some remarks/suggestions though!

imod/tests/test_mf6/test_mf6_wel.py
```diff
@@ -137,35 +141,141 @@ def test_to_mf6_pkg__high_lvl_transient(basic_dis, well_high_lvl_test_data_trans
     np.testing.assert_equal(mf6_ds["rate"].values, rate_expected)
 
 
-def test_clip_box__high_lvl_stationary(well_high_lvl_test_data_stationary):
+class ClipBoxCases:
```
JoerivanEngelen, line 144: Looks better! I'm still missing a test with an unstructured grid, you can easily convert a structured grid to an unstructured grid with: 
``idomain_ugrid = xu.UgridDataArray.from_structured(idomain)``

imod/mf6/wel.py
Manangka: This worked like a charm. I added 2 unit tests for unstructured grids and with your code they both pass

Issue #629, opened 2023-11-07
The simulationl regrid() and clip () should remove packages that are invalid after regridding.


Manangka:
In GitLab by @HendrikKok on Nov 7, 2023, 10:45

For example: a RIV-package could become filled with only nan’s after regridding to a smaller domain.

Simulation regrid and clip should pop empty packages. We could use a warning returned by dask for this (saves performance?), but more transparent would be to check if all cells are nan.

PR #742, opened 2024-01-12
Issue #629 purge empty packages

luitjansl:
Solves story #629. This PR adds a function to package_base that returns whether the package is empty or not. This in turn is used in a second new function that removes empty packages from a model.

### DIFF
--- a/imod/mf6/hfb.py
+++ b/imod/mf6/hfb.py
@@ -403,6 +403,14 @@ class HorizontalFlowBarrierBase(BoundaryCondition, ILineDataPackage):
 
         return Mf6HorizontalFlowBarrier(**barrier_dataset)
 
+    def is_empty(self) -> bool:
+        if super(__class__, self).is_empty():
+            return True
+
+        linestrings = self.dataset["geometry"]
+        only_empty_lines = all([ls.is_empty for ls in linestrings.values])
+        return only_empty_lines
+
     def _resistance_layer(
         self,
         snapped_dataset: xu.UgridDataset,
--- a/imod/mf6/package.py
+++ b/imod/mf6/package.py
@@ -283,10 +283,16 @@ class Package(PackageBase, abc.ABC):
         """
         Returns True if the package is empty- for example if it contains only no-data values.
         """
+
+        # Create schemata dict only containing the
+        # variables with a AllNoDataSchema and EmptyIndexesSchema (in case of
+        # HFB) in the write schemata.
         allnodata_schemata = filter_schemata_dict(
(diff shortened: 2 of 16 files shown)


imod/mf6/hfb.py
```diff
@@ -403,6 +403,14 @@ def to_mf6_pkg(
 
         return Mf6HorizontalFlowBarrier(**barrier_dataset)
 
+    def is_empty(self) -> bool:
+        if super(__class__, self).is_empty():
```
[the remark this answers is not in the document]
luitjansl: if the hfb object has linestrings, then the baseclass.is_empty() will return False, because the Geometry object is in there, and it conforms the EmptyIndexesSchema. 
That's why after asking the base class, we check if all the linestrings are empty. If they are, then hfb.is_empty()  returns True.

imod/mf6/package.py
```diff
@@ -278,6 +279,17 @@ def _validate(self, schemata: Dict, **kwargs) -> Dict[str, List[ValidationError]
                         errors[variable].append(e)
         return errors
 
+    def is_empty(self) -> bool:
+        """
+        Returns True if the package is empty- for example if it contains only no-data values.
+        """
+        allnodata_schemata = filter_schemata_dict(
```
JoerivanEngelen, line 286: I think this comment is still useful:

```python
        # Create schemata dict only containing the
        # variables with a AllNoDataSchema and EmptyIndexesSchema (in case of
        # HFB) in the write schemata.
```

imod/mf6/package.py
```diff
@@ -278,6 +279,17 @@ def _validate(self, schemata: Dict, **kwargs) -> Dict[str, List[ValidationError]
...
+            self._write_schemata, (AllNoDataSchema, EmptyIndexesSchema)
+        )
+
+        allnodata_errors = self._validate(allnodata_schemata)
```
JoerivanEngelen, line 290: This might be useful to add:

```python
        # Find if packages throws ValidationError for AllNoDataSchema or
        # EmptyIndexesSchema.
```

JoerivanEngelen (approved):
Looks good, approved. Good idea to move this to a method. I see some comments got removed in the migration; perhaps a part of that is still useful to move as well?

imod/mf6/package.py
luitjansl: done

imod/mf6/package.py
luitjansl: done

imod/mf6/package.py
luitjansl: done

PR #755, opened 2024-01-19
Issue #752 Add documentation for the logger to the API reference

Manangka:
In this commit the documentation for the logger is added to the api reference.

To make sure that any further changes to the logger are automatically added to the api reference 2 templates have been added. These templates can be used with the autosummary recursive option.  The result being that documentation is generated for  all modules, classes, methods and functions defined in  a root module. The exception being private ones (the ones starting with `_` or `__`)

Note:
The base branch needs to be changed to master after #754 has been merged

### DIFF
--- a/docs/_templates/custom-class-template.rst
+++ b/docs/_templates/custom-class-template.rst
@@ -1,0 +1,32 @@
+{{ fullname | escape | underline}}
+
+.. currentmodule:: {{ module }}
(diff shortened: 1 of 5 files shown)


luitjansl (approved):
looks good, no comments

JoerivanEngelen (approved):

PR #760, opened 2024-01-19
Use hatchling instead of setuptools, add pixi task: pypi-publish

Huite:
This setups the pyproject to use hatching instead of setuptools.

It also adds a pypi-publish task to pixi to make the release process a little smoother.

Finally, the pyproject version is derived from the imod.__init__. Ideally, there's a single source of truth, but pixi still contains its own version number (note: that one is not used anywhere, you could set it to a dummy value as well).


Huite:
I just set this up for xugrid, and it seems to be working.

Keep in mind that you will need to update the conda-recipe for the host requirements. See for example:
https://statics.teams.cdn.office.net/evergreen-assets/safelinks/1/atp-safelinks.html

pixi.toml
```diff
@@ -38,6 +38,7 @@ examples = { cmd = [
     "--verbose",
     "--junitxml=examples_report.xml",
 ], depends_on = ["install"], cwd = "imod/tests" }
+pypi-publish = { cmd = "rm --recursive --force dist && python -m build && twine check dist/*" }
```
Hofer-Julian, line 41: Maybe add `hatchling` to the dependencies?

Hofer-Julian (commented):
Apart from the one comment looks good to me

Hofer-Julian:
> FYI: The developer also promotes `Hatch`, which would kinda replace `pixi`, but I don't think that would work for us.

Agreed. 
`hatchlings` on its own mostly helps us by requiring less configuration and that it makes IDEs support editable installs properly

pixi.toml
Huite: Good catch. I've added it to the pyproject dev dependencies.

Huite:
I removed the MANIFEST.in, which is no longer needed, and I re-added the upload command to the new pixi task.
To test it, just remove the upload part (or cancel when asked to submit your pypi name/token/password).

pixi.toml
Huite: By the way, we could arguably remove the dev dependencies since we recommend pixi anyway, right?

And hatchling then wouldn't be need for a regular user.

Hofer-Julian: Yeah, that should be fine.

Hofer-Julian: Still annoying that there is no good way to find out if you require too much or especially too little in your normal dependencies.

JoerivanEngelen (approved):
Looks good to me, just a question: Does this require developers to do a new editable install on their machines?

Huite:
I don't think so, and otherwise you just run pixi run install...

Hofer-Julian:
> Looks good to me, just a question: Does this require developers to do a new editable install on their machines?

Good that you mention that.
Yesterday I saw that editable installations had to be redone on @SouthEndMusic's machine, so I think the answer is yes.

JoerivanEngelen:
> > Looks good to me, just a question: Does this require developers to do a new editable install on their machines?
> 
> Good that you mention that. Yesterday I saw that editable installations had to be redone on @SouthEndMusic's machine, so I think the answer is yes.

Ah good to know, this means the advice has to be added to the changelog: _"Users who did a development install are adviced to run ``pip uninstall imod`` followed by ``pip install -e .`` again"_. I'll add that, and I'll fix the conflict in .lock files.

JoerivanEngelen:
We shouldn't forget to update the imod-feedstock for the next release, I made an issue to remind myself here: https://github.com/conda-forge/imod-feedstock/issues/31

JoerivanEngelen:
Blocked by: https://github.com/Deltares/imod-python/issues/774, wait until https://github.com/Deltares/imod-python/pull/775 is merged.

PR #762, opened 2024-01-22
IIssue # 761 Replace isort with ruff sort

Manangka:
As of lately isort became rather slow taking between 1-3 minutes to run.
This commit replaces isort with ruff sort which basically works instantaneously


JoerivanEngelen (approved):
Looks good!

PR #776, opened 2024-01-24
Issue #764 modelsplit auxiliary variables

### LINKED ISSUE #764: Check if aux vars containing concentrations preserved in GWF packages upon splitting
DoD: If not present, add functionality. If no changes required, task is done.

luitjansl:
Fixes #764

# Description
Allows packages with auxiliary data (such as flow boundary conditions when used with transport) to be split like other packages.
Also added a test.


imod/mf6/auxiliary_variables.py
```diff
@@ -25,11 +25,40 @@ def _get_auxiliary_data_variable_names_mapping(package):
     return result
 
 
-def add_periodic_auxiliary_variable(package):
+def add_periodic_auxiliary_variable(package) -> None:
```
luitjansl, line 28: due to circular import, the Pakcage class cannot be loaded in this source file, and no type information can be provided on package.

imod/tests/fixtures/flow_transport_simulation_fixture.py
```diff
@@ -148,7 +177,7 @@ def flow_transport_simulation():
         gwf_model, "species_d", 10.0, 5.0, 0.002
     )
 
-    simulation["flow_solver"] = imod.mf6.Solution(
+    simulation["solver"] = imod.mf6.Solution(
```
luitjansl, line 180: the string "flow" is hard-coded somewhere, so if  this Solution is not called "solver" bad things happen.

imod/mf6/auxiliary_variables.py
Manangka: This would be a good time to introduce the IPackage interface to break the circular dependency. We already have the IPackageBase in the interface folder.

The interface would be:

```
class IPackage(IPackageBase):
    """
    The base methods and attributes available in all packages
    """

    @property
    @abstractmethod
    def auxiliary_data(self) -> Dict[str, str]:
        raise NotImplementedError
```


In the Package class:

```
class Package(PackageBase, IPackage):

....

@property
def auxiliary_data(self) -> Dict[str, str]:
    if (hasattr(self, "_auxiliary_data")):
        return self._auxiliary_data
    else: 
        return {}
```


In the auxiliarry_variables file:
```
def get_variable_names(package: IPackage):
    ...

```

imod/mf6/auxiliary_variables.py
```diff
@@ -25,11 +25,40 @@ def _get_auxiliary_data_variable_names_mapping(package):
...
+    """
+    splits an auxiliary dataarray (with one or more auxiliary variable dimension) into dataarrays per
```
Manangka, line 30: Splits should start with a capital

imod/mf6/auxiliary_variables.py
```diff
@@ -25,11 +25,40 @@ def _get_auxiliary_data_variable_names_mapping(package):
...
+    auxiliary variable dimension. For example a concentration auxiliary variable in a flow package
+    will have a species dimension, and will be split in several dataarrays- one for each species.
+    """
     if hasattr(package, "_auxiliary_data"):
```
Manangka, line 34: with the interface and property in place as described above you don't need this if statement anymore and can directly iterate over 

`for aux_var_name, aux_var_dimensions in package.auxiliary_data.items():`

This also applied to the method below

imod/mf6/auxiliary_variables.py
```diff
@@ -25,11 +25,40 @@ def _get_auxiliary_data_variable_names_mapping(package):
...
             for s in aux_coords:
                 package.dataset[s] = package.dataset[aux_var_name].sel(
                     {aux_var_dimensions: s}
                 )
+
+
+def remove_periodic_auxiliary_variable(package) -> None:
+    """
+    removes the data arrays created by add_periodic_auxiliary_variable(...) but does not
+    remove the auxiliary dataarray used as source for add_periodic_auxiliary_variable(...)
```
Manangka, line 46: Use the following form when refering to another method to automatically generate a link in sphinx:
`:meth:`imod.mf6.auxiliary_variables.add_periodic_auxiliary_variable``

imod/mf6/auxiliary_variables.py
```diff
@@ -25,11 +25,40 @@ def _get_auxiliary_data_variable_names_mapping(package):
...
             aux_coords = package.dataset[aux_var_name].coords[aux_var_dimensions].values
             for s in aux_coords:
                 package.dataset[s] = package.dataset[aux_var_name].sel(
                     {aux_var_dimensions: s}
                 )
+
+
+def remove_periodic_auxiliary_variable(package) -> None:
+    """
+    removes the data arrays created by add_periodic_auxiliary_variable(...) but does not
```
Manangka, line 45: Start the sentences of your docstring with a capital

imod/mf6/auxiliary_variables.py
```diff
@@ -25,11 +25,40 @@ def _get_auxiliary_data_variable_names_mapping(package):
...
+        for aux_var_name, aux_var_dimensions in package._auxiliary_data.items():
+            if aux_var_dimensions in package.dataset.coords:
+                for species in package.dataset.coords[aux_var_dimensions].values:
+                    if species in list(package.dataset.keys()):
+                        package.dataset = package.dataset.drop_vars(species)
+
+
+def has_auxiliary_variable_source_array(package) -> bool:
+    """
+    returns True if a package contains an auxiliary data source data array
```
Manangka, line 58: Start with a capital

imod/mf6/auxiliary_variables.py
```diff
@@ -25,11 +25,40 @@ def _get_auxiliary_data_variable_names_mapping(package):
...
     if hasattr(package, "_auxiliary_data"):
         for aux_var_name, aux_var_dimensions in package._auxiliary_data.items():
             aux_coords = package.dataset[aux_var_name].coords[aux_var_dimensions].values
             for s in aux_coords:
                 package.dataset[s] = package.dataset[aux_var_name].sel(
                     {aux_var_dimensions: s}
                 )
+
+
+def remove_periodic_auxiliary_variable(package) -> None:
```
Manangka, line 43: What about the following method names?
```
add_periodic_auxiliary_variable -> add_periodic_auxiliary_variable_to_dataset
remove_periodic_auxiliary_variable -> remove_periodic_auxiliary_variable_from_dataset
has_auxiliary_variable_source_array -> has_auxiliary_variable_in_data_set
```

Anyhow the term `source array `is confusing. I think `dataset` or `package dataset`  is clearer

imod/mf6/multimodel/modelsplitter.py
```diff
@@ -86,6 +92,10 @@ def slice_model(partition_info: PartitionInfo, model: Modflow6Model) -> Modflow6
         new_idomain = model.domain.sel(coords).where(sliced_domain_2D, other=0)
 
     for pkg_name, package in model.items():
+        if isinstance(package, BoundaryCondition):
```
Manangka, line 95: Shouldn't this apply to all packages, not just the BoundaryCondition packages?

Manangka (changes requested):

imod/mf6/multimodel/modelsplitter.py
luitjansl: The boundary conditions are special, because over a boundary condition water will enter or leave the model. Usually we assign a concentration to the boundary condition, which is used when water is entering the model over that boundary. When water leaves the model, it leaves with the concentration computed in the model. That's why the flow boundary conditions carry this concentration data if the flow model is used with transport. But the other packages (STO, NPF, DIS, etcetera) do not carry concentration data, and therefore don't need this operation.

imod/mf6/auxiliary_variables.py
luitjansl: done

imod/mf6/auxiliary_variables.py
luitjansl: done

imod/mf6/auxiliary_variables.py
luitjansl: done

imod/mf6/multimodel/modelsplitter.py
Manangka: Isn't there a reason for those packages to have auxiliary variables for other reasons? Like the “ANGLDEGX” and “CDIST” auxiliary variables. Are those auxiliary variables propogates aswell? Or isnt that necesarry?

imod/mf6/auxiliary_variables.py
luitjansl: renamed them. I removed the "periodic"part because these auxiliary variables contain data that is potentially time dependent, but usually not periodic. 
I now call them 
add_periodic_auxiliary_variable -> expand_auxiliary_variables
remove_periodic_auxiliary_variable ->remove_expanded_auxiliary_variables_from_dataset
has_auxiliary_variable_source_array -> removed

imod/mf6/multimodel/modelsplitter.py
JoerivanEngelen: Important here is: remove_**periodic**_auxiliary_variable, so these are aux variables which can change over time (if I'm not mistaken). This excludes the aux vars for the exchange packages. We might need to clarify that more in the method/function names.

imod/tests/fixtures/flow_transport_simulation_fixture.py
JoerivanEngelen: The key ``"flow"`` is already used to assign the groundwater model in this example (line 166):

```python
simulation["flow"] = gwf_model
```

However the fact that the key should be "solver" is worrysome. 
Perhaps this is worthy an issue by itself? A quick CTRL+F search shows that the "solver" key is hardcoded in the ``split`` method, which is wrong. We can use the ``_pkg_id`` for that.

imod/tests/test_mf6/test_multimodel/test_mf6_modelsplitter.py
```diff
@@ -71,3 +72,33 @@ def test_slice_model_unstructured(circle_model):
...
+
+
+@pytest.mark.usefixtures("flow_transport_simulation")
+def test_slice_model_with_auxiliary_variables(tmp_path, flow_transport_simulation):
+    flow_simulation = flow_transport_simulation
+    flow_simulation.pop("tpt_a")
+    flow_simulation.pop("tpt_b")
+    flow_simulation.pop("tpt_c")
+    flow_simulation.pop("tpt_d")
+    flow_simulation.pop("transport_solver")
```
JoerivanEngelen, line 84: Note: You are mutating a fixture which has "session" as scope, meaning this results in unwarranted side-effects for other tests, which happen after this. I think you should alter the scope of the fixture to ``"function"``

JoerivanEngelen (changes requested):

imod/tests/test_mf6/test_multimodel/test_mf6_modelsplitter.py
luitjansl: done

imod/mf6/auxiliary_variables.py
luitjansl: done

imod/tests/fixtures/flow_transport_simulation_fixture.py
luitjansl: as discussed, @JoerivanEngelen created an issue for this: #778

imod/mf6/multimodel/modelsplitter.py
luitjansl: indeed, and in this PR I renamed _periodic_ to _transient_, because periodic somehow suggests a repeating pattern in the data.

JoerivanEngelen: Good!

JoerivanEngelen (approved):

Manangka (approved):

Issue #737, opened 2024-01-10
Add Github template for pull requests


JoerivanEngelen:
It would be helpful to introduce a template for PRs with a checklist of things to consider/include. This would help avoiding some common mistakes developers make (including myself). For example in about half of the merged PRs, we forget to update the changelog with relevant info for the user, which then has to be added later.

Things to include in the the checklist/definition of done:

- Link to issue
- Description of changes broadly made and how they are going to fix the issue
- Update changelog, if changes are relevant to the users (e.g. bugfix, API changes, changes in behaviour)
- PR title starts with issue number, like ``#737``

PR #777, opened 2024-01-25
Issue #757 Create nightly pipeline

### LINKED ISSUE #757: separate pipeline for modflow's nightly build ?
sometimes we are interested in using features or fixes only present in the nightly build of MF6 (for example to obtain correct CHD fluxes at partition boundaries).
We should be able to use it for testing purposes in the pipeline, but currently some transport tests fail with it. 

We should address this, and  in such a manner that we don't loose support for the last stable release. Perhaps a separate pipeline for each MF6 version?

Manangka:
Fixes #757

# Description
This commit adds a nightly pipeline.
The goal of the nightly pipeline is to build or repository against the latest development branch of MODFLOW.
This will allow us to catch breaking changes early on.

The pipeline file has been refactored to prevent code duplication. This has been accomplished by using templates.
The new nightly pipeline is added as a new subproject of the main project

# Checklist

- [x] Links to correct issue
- [ ] Unit tests were added

### DIFF
--- a/.teamcity/settings.kts
+++ b/.teamcity/settings.kts
@@ -70,6 +70,8 @@ object Examples : BuildType({
     vcs {
         root(DslContext.settingsRoot, "+:. => imod-python")
         root(AbsoluteId("iMOD6_IMODPython_MetaSwapLookupTable"), ". => lookup_table")
+
+        cleanCheckout = true
     }
 
     steps {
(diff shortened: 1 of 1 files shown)


luitjansl (commented):
I cannot review this.

JoerivanEngelen (approved):
Hard to verify from xml as a n00b, so I assume it works fine ;-)

PR #781, opened 2024-01-26
Issue #746 slice model transport

### LINKED ISSUE #746: Add slice_model support GroundwaterTransport model
``imod.mf6.multimodel.modelsplitter.slice_model`` now has ``GroundwaterFlowModel`` hardcoded to create new models. This should also be able to support GroundwaterTransport, based on the type of the provided model.

NOTE: It might be that issues with transport packages might pop up when this is done. If too many issues show up, best is to create a separate issue for those issues.

luitjansl:
Fixes #746

# Description
Fixes the slice_model function to slice transport models as well. 
Includes changes to the Advection package. The advection scheme parameter is set by the constructors of derived types of advection. Trying to reinitialize the derived type throws an error, as its construction method doesn't expect arguments. As a pragmatic solution, the constructor methods now get a ``**ignored`` argument to ignore these kwargs.

- [x] Links to correct issue
- [x] Unit tests were added


JoerivanEngelen (approved):
Looks good

Manangka:
In your description you mentioned that you get an error when derived class of the Advection package is bein reinitialized. In your code changes i don't see where this is happening. Could you tell me where this object is being reinitialized?

luitjansl:
> In your description you mentioned that you get an error when derived class of the Advection package is bein reinitialized. In your code changes i don't see where this is happening. Could you tell me where this object is being reinitialized?

In package.py, in the function "mask", last line

Manangka:
> > In your description you mentioned that you get an error when derived class of the Advection package is bein reinitialized. In your code changes i don't see where this is happening. Could you tell me where this object is being reinitialized?
> 
> In package.py, in the function "mask", last line

Instead of adding the ignore parameter you could also override the mask method in the Advection package. You could do the same as in the HFB package where the mask method returns a deepcopy

luitjansl:
> > > In your description you mentioned that you get an error when derived class of the Advection package is bein reinitialized. In your code changes i don't see where this is happening. Could you tell me where this object is being reinitialized?
> > 
> > 
> > In package.py, in the function "mask", last line
> 
> Instead of adding the ignore parameter you could also override the mask method in the Advection package. You could do the same as in the HFB package where the mask method returns a deepcopy

done

PR #804, opened 2024-02-01
Issue #795 Add pipeline which periodically updates the projects dependencies to the latest version

### LINKED ISSUE #795: Create Pixi dependencies update pipeline

Manangka:
Fixes #795

This commit adds a pipeline that periodically updates the project dependencies.
It runs every sunday at 15:00. When changes are found a new update branch is pushed and a PR is created.

The created PR in turn will trigger the normal build pipeline so all the regular checks are run against this branch.
If any of the tests fail they should be fixes as part of the created PR

Reviewers are automatically assigned in this case that are @luitjansl and @JoerivanEngelen 

# Checklist

- [x] Links to correct issue
- [ ] Unit tests were added

### DIFF
--- a/.teamcity/settings.kts
+++ b/.teamcity/settings.kts
@@ -7,6 +7,7 @@ import jetbrains.buildServer.configs.kotlin.buildSteps.script
 import jetbrains.buildServer.configs.kotlin.projectFeatures.ProjectReportTab
 import jetbrains.buildServer.configs.kotlin.projectFeatures.projectReportTab
 import jetbrains.buildServer.configs.kotlin.triggers.finishBuildTrigger
+import jetbrains.buildServer.configs.kotlin.triggers.schedule
 import jetbrains.buildServer.configs.kotlin.triggers.vcs
 
 /*
@@ -66,6 +67,7 @@ project {
     }
 
     subProject(Nightly)
+    subProject(Pixi)
 }
 
 object GitHubIntegrationTemplate : Template({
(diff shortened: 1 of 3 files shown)


Manangka:
@Hofer-Julian You might find this interesting as well

.teamcity/settings.kts
```diff
@@ -422,4 +424,89 @@ object NightlyTests : BuildType({
...
+
+    steps {
+        powerShell {
+            name = "Download Github CLI"
+            id = "Download_Github_CLI"
+            formatStderrAsError = true
+            scriptMode = script {
+                content = """
+                    Invoke-WebRequest -Uri  https://github.com/cli/cli/releases/download/v2.42.1/gh_2.42.1_windows_amd64.zip -OutFile gh.zip
+                    Expand-Archive gh.zip -DestinationPath gh
```
Hofer-Julian, line 457: You can just add `gh` as pixi dependency.
Alternative it's also available as module on linux called `github`.
I would suggest the former though.

.teamcity/settings.kts
```diff
@@ -422,4 +424,89 @@ object NightlyTests : BuildType({
...
+            name = "Update dependencies"
+            id = "Update_dependencies"
+            workingDir = "imod-python"
+            scriptMode = script {
+                content = """
+                    ${'$'}env:Path = '%system.teamcity.build.checkoutDir%\gh\bin;' + ${'$'}env:Path  
+                    
+                    echo "Create update branch"
+                    
+                    git remote set-url origin https://%GT_USER%:%env.GH_TOKEN%@github.com/Deltares/imod-python.git
```
Hofer-Julian, line 471: Why is that necessary?

Manangka: The credentials used during checkout aren't strored. This is by design by TeamCity for security purposes.
Therefor we have to reapply the user and token to be able to commit to the branch

.teamcity/settings.kts
Manangka: I'll be going for your first suggestion. Mainly because I don't want to introduce a new scripting language besides the ones that already being used

luitjansl (approved):

PR #819, opened 2024-02-05
Issue #798 partition ssm packages

### LINKED ISSUE #798: update ssm package when splitting flow + transport simulations
The ssm package is part of a transport model, and it  describes what flow boundaries to take into account for transport. For example:
```
begin options
  save_flows
end options

begin sources
  chd aux species_b
  rch aux species_b
  well aux species_b
end sources
```
But when splitting into subdomains, not every flow boundary will be present in every partition, and this will cause modflow to fail to run the simulation

luitjansl:
Fixes #798 

When we split a simulation of both flow and transport models, the ssm package of the transport simulation is re-generated after splitting, because not every flow boundary will be present in every partition. 

- [X] Links to correct issue
- [X] Unit tests were added


imod/mf6/ssm.py
```diff
@@ -146,7 +146,7 @@ def from_flow_model(
                 aux_var_names.append(species)
 
         if len(names) == 0:
-            raise ValueError("flow model does not contain boundary conditions")
```
JoerivanEngelen, line 149: Is this appropriate behavior? Or required?
We can log a warning?

JoerivanEngelen (commented):
One comment!

imod/mf6/ssm.py
luitjansl: before we implemented partitioning, this was a red flag. But now, we can have a partition which is not on the boundary, and it will not have flow boundaries,  so it is not a red flag anymore.

imod/mf6/ssm.py
```diff
@@ -146,7 +149,10 @@ def from_flow_model(
...
+            if is_split:
+                logger.info("generating ssm package without exchanges")
+            else:
+                raise ValueError("flow model does not contain boundary conditions")
```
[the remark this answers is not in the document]
luitjansl: done, committed the suggestion and added return statement

JoerivanEngelen (approved):

Issue #563, opened 2023-10-04
Fix MyPy errors in mf6 module


Manangka:
At the moment MyPy reports more than 100 issues in the imod/mf6 folder. We need to resolve them and change the pipeline from giving a warning to giving an error when new MyPy errors are reported

Manangka:
In GitLab by @JoerivanEngelen on Oct 26, 2023, 10:55

marked this issue as related to #65

JoerivanEngelen:
Some work on this has been done in: https://github.com/Deltares/imod-python/issues/681, https://github.com/Deltares/imod-python/issues/682, https://github.com/Deltares/imod-python/issues/684

JoerivanEngelen:
Issue description referred to mf6 module, is fixed now.

PR #837, opened 2024-02-12
Issue #836 Make MyPy stricter by enabling follow_imports

### LINKED ISSUE #836: Make MyPy stricter. Enable follow_imports

Manangka:
Fixes #836

In this commit the settings of MyPy are made stricter by enabling the follow_imports flag. By enabling this flag, information about type hints from modules being imported are also taken into account when performing the static code analyzes

In the pipeline a threshold has been added to the MyPy step which tolerates 2 errors. These 2 errors have to do with mypy errors with regard to the regridder. Those errors are better handled in a separate story. Once resolved the threshold can be set to 0

# Checklist

- [x] Links to correct issue
- [ ] Unit tests were added

### DIFF
--- a/imod/mf6/npf.py
+++ b/imod/mf6/npf.py
@@ -19,7 +19,9 @@ def _dataarray_to_bool(griddataarray: GridDataArray) -> bool:
         return False
 
     if griddataarray.values.size != 1:
-        raise ValueError("Dataarray is not a single value. Unable to convert to boolean")
+        raise ValueError(
+            "Dataarray is not a single value. Unable to convert to boolean"
+        )
     return griddataarray.values.item()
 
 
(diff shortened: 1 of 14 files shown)


imod/mf6/npf.py
```diff
@@ -11,6 +11,18 @@
...
+
+def _dataarray_to_bool(griddataarray: GridDataArray) -> bool:
+    if griddataarray is None or griddataarray.values is None:
+        return False
+
+    if griddataarray.values.size != 1:
+        raise ValueError(
+            "Dataarray is not a single value. Unable to convert to boolean"
+        )
+    return griddataarray.values.item()
```
luitjansl, line 25: you could add a check on the dtype of griddataarray. It seems like in the use cases, it should be an actual bool and not an integer or float (and if it is, there may be something wrong)

luitjansl (approved):
looks good!

imod/mf6/npf.py
Manangka: Good point. Added the check

JoerivanEngelen:
Related to: https://github.com/Deltares/imod-python/issues/563

PR #840, opened 2024-02-12
Issue #627 unclear docstring

### LINKED ISSUE #627: Unclear docstring in Well-class on variable ‘rate’ for transient cases
In GitLab by @HendrikKok on Nov 7, 2023, 10:28

This should not be a list of floats, but dataset with index and time?

luitjansl:
Fixes #627 

# Description
added xr.DataArray as a possible type for the rate variable of the high level well package.

# Checklist

- [X] Links to correct issue
- [X] Unit tests were added


JoerivanEngelen (changes requested):
The changes made are an improvement, but they don't solve this issue. The issue pertained to the docstrings (lines 89-91), which remain unchanged with the current changeset. 

I think we need to the following:
```
    rate: float, list of floats or xr.DataArray
        is the volumetric well rate. A positive value indicates well
        (injection) and a negative value indicates discharge (extraction) (q).
        If provided as DataArray, an ``"index"`` dimension is required and an
        optional ``"time"`` dimension and coordinate specify transient input. 
        In the latter case, it is important that dimensions are in the order:
        ``("time", "index")``
```

Below we can add an example:

```
    Examples
    ---------

    >>> screen_top = [0.0, 0.0]
    >>> screen_bottom = [-2.0, -2.0]
    >>> y = [83.0, 77.0]
    >>> x = [81.0, 82.0]
    >>> rate = [1.0, 1.0]

    >>> imod.mf6.Well(x, y, screen_top, screen_bottom, rate)

    For a transient well:
    
    >>> weltimes = pd.date_range("2000-01-01", "2000-01-03")

    >>> rate_factor_time = xr.DataArray([0.5, 1.0], coords={"time": weltimes}, dims=("time",))
    >>> rate_transient = rate_factor_time * xr.DataArray(rate, dims=("index",))

    >>> imod.mf6.Well(x, y, screen_top, screen_bottom, rate_transient)

```

I also see the the order of parameters in the docstrings doesn't follow the actual order of arguments. Could you adapt the docstrings so that they follow the order: x, y, screen_top, screen_bottom?

JoerivanEngelen (approved):
Looks good!

Manangka (approved):

imod/mf6/wel.py
```diff
@@ -207,7 +207,7 @@ def __init__(
         repeat_stress: Optional[xr.DataArray] = None,
     ):
         if id is None:
-            id = np.arange(len(x)).astype(str)
+            id = list(np.arange(len(x)).astype(str))
```
Huite, line 210: Nitpick, in this case I'd suggest:
`ids = [str(i) for i in range(len(x))]`

`id` is also a built-in function of Python that you shouldn't redefine.

luitjansl: done, now using `[str(i) for i in range(len(x))]`. 
`id` is the name of a constructor argument, so I'm hesitant to rename it as it may break user scripts.

Huite: That's indeed a good argument to keep `id` (it's not like we especially need the `id` in this function's scope).

PR #842, opened 2024-02-13
Issue #839 2 species transport

### LINKED ISSUE #839: add test with balance output and at least 2 species controlling density
Add  a test that splits a flow + transport simulation in subdomains. There should be at least 2 species controlling density. 
At least one of the partitions should not be at the model boundary, but internal. The transport model should compute balance, and merged balance results should be compared. 

Note (20-2-'24):
The scope of this issue has been reduced. Balance output checking will be done in a follow up issue. This issue now only focusses on making a multispecies model with variable density, partitioning it  and checking concentration results.

luitjansl:
Fixes #839 

# Description
partitions a flow +  multispecies transport model. Implements an update for the buyoancy package, now it checks that the order in which new names are assigned is correct ( it assumes that the new names are the old names + a suffix ) 
Includes a fix needed for loading transport balances. 

# Checklist

- [X] Links to correct issue
- [X] Unit tests were added


imod/mf6/buy.py
```diff
@@ -157,6 +157,9 @@ def update_transport_models(self, new_modelnames: Sequence[str]):
         transport_model_names = self.get_transport_model_names()
         if not len(transport_model_names) == len(new_modelnames):
             raise ValueError("the number of transport models cannot be changed.")
+        for i in range (0, len(transport_model_names)):
```
JoerivanEngelen, line 160: Can be simplified to:

```suggestion
        for i in range (len(transport_model_names)):
```

Or even better:

```suggestion
        for modelname, new_modelname in zip(transport_model_names, new_modelnames):
            if not modelname in new_modelname:
```

imod/tests/fixtures/mf6_circle_fixture.py
```diff
@@ -340,3 +340,85 @@ def circle_model_transport():
...
+            xt3d_off=False,
+            xt3d_rhs=False,
+        )
+        transport_model["adv"] = imod.mf6.AdvectionUpstream()
+        transport_model["mst"] = imod.mf6.MobileStorageTransfer(porosity, save_flows=True)
+
+        # %%
+        # Now we define initial conditions (0.0) and output options for the transport
+        # simulation, and assign the transport model to the simulation as well.
+        transport_model["ic"] = imod.mf6.InitialConditions(start=max_concentration)
```
JoerivanEngelen, line 390: Comment doesn't match argument provided to ``start``

imod/tests/fixtures/mf6_circle_fixture.py
```diff
@@ -340,3 +340,85 @@ def circle_model_transport():
...
+@pytest.fixture(scope="function")
+def circle_model_transport_multispecies():
+
+    al = 0.001
+    porosity = 0.3
+    max_concentration = 35.0
+    min_concentration = 0.0
+    max_density = 1025.0
+    min_density = 1000.0
+    species = ["sali", "tempe"]
```
JoerivanEngelen, line 355: Maybe this is a bit better:

```suggestion
    species = ["salt", "temp"]
```

imod/tests/fixtures/mf6_circle_fixture.py
```diff
@@ -340,3 +340,85 @@ def circle_model_transport():
...
+        )
+
+        simulation[f"tpt_{specie}"] = transport_model
+    slope = (max_density - min_density) / (max_concentration - min_concentration)    
+    modelnames = [f"tpt_{specie}" for specie in species]    
+    gwf_model["buoyancy"] = imod.mf6.Buoyancy(
+        reference_density=min_density,
+        modelname=modelnames,
+        reference_concentration=[min_concentration,min_concentration ],
+        density_concentration_slope=[slope/2, slope/2],
```
JoerivanEngelen, line 402: Why the division by 2?

JoerivanEngelen (changes requested):
Looks good, thanks also for the extra janitor work (e.g. type annotation fixes). I have a few suggestions!

imod/tests/fixtures/mf6_circle_fixture.py
luitjansl: the original model had 1 species, now we have 2 that have the same initial conditions and boundaries. So if I were to use the same slope, the water would get 2 times heavier. 
But indeed that is not an issue perse, so now I  removed the "/2"

examples/mf6/circle_transport.py
```diff
@@ -250,9 +250,9 @@
 transport_model["adv"] = imod.mf6.AdvectionUpstream()
 transport_model["mst"] = imod.mf6.MobileStorageTransfer(porosity)
 
-# %%
-# Now we define initial conditions (0.0) and output options for the transport
-# simulation, and assign the transport model to the simulation as well.
+# %% Define the maximum concentration as the initial conditions, also output
+# options for the transport model, and assign the transport model to the
```
JoerivanEngelen, line 254: I think the sphinx converter to Jupyter blocks requires a newline after the ``# %%`` to render properly.

```suggestion
# %%
# Define the maximum concentration as the initial conditions, also output
# options for the transport model, and assign the transport model to the
```

imod/tests/fixtures/mf6_circle_fixture.py
```diff
@@ -310,9 +310,9 @@ def circle_model_transport():
     transport_model["adv"] = imod.mf6.AdvectionUpstream()
     transport_model["mst"] = imod.mf6.MobileStorageTransfer(porosity)
 
-    # %%
-    # Now we define initial conditions (0.0) and output options for the transport
-    # simulation, and assign the transport model to the simulation as well.
+    # %% Define the maximum concentration as the initial conditions, also output
+    # options for the transport model, and assign the transport model to the
+    # simulation as well.
```
JoerivanEngelen, line 315: See previous comment (less important because this is not rendered as html):

```suggestion
    # %% 
    # Define the maximum concentration as the initial conditions, also output
    # options for the transport model, and assign the transport model to the
    # simulation as well.
```

JoerivanEngelen (approved):
One minor comment to get the example rendered properly.

imod/mf6/model_gwf.py
```diff
@@ -143,7 +143,7 @@ def update_buoyancy_package(self, tpt_models_per_flow_model)->None:
             return
         buoyancy_package = self[buoyancy_key]
         tpt_models_old = buoyancy_package.get_transport_model_names()
```
Manangka, line 145: I wouldn't abbreviate tpt. Instead i would write transport_models

Manangka:
Could you update the threshold in the pipeline aswell. Now that all mypy errors are fixed we can set it to 0.
You can find the pipeline in  .teamcity/settings.kts. If you search for `threshold` you should get 2 hits both holding the value 2.
Please change both to 0

imod/mf6/model_gwf.py
luitjansl: done

PR #843, opened 2024-02-14
Issue #679 fix split and dump for partitioned models

### LINKED ISSUE #679: Split simulation cannot be dumped
In GitLab by @JoerivanEngelen on Dec 5, 2023, 16:26

To reproduce:

- Split simulation
- Dump splitted simulation

Throws:

```
---------------------------------------------------------------------------
AttributeError                            Traceback (most recent call last)
c:\Users\engelen\projects_wdir\alameda_creek\build-model.py in line 3
      187 # %% write model files
----> 189 simulation_split.dump(modeldir_out / "partitioned")
      191 # simulation_split.write(modeldir_out + "/" + name, binary = False)#, validate = False)

File C:\src\imod-python\imod\mf6\simulation.py:699, in Modflow6Simulation.dump(self, directory, validate, mdal_compliant)
    697     else:
    698         path = f"{key}.nc"
--> 699         value.dataset.to_netcdf(directory / path)
    700         toml_content[cls_name][key] = path
    702 with open(directory / f"{self.name}.toml", "wb") as f:

AttributeError: 'list' object has no attribute 'dataset'
```

This is because exchanges are saved in a list. The ``write`` method has this part:

```python
    def write(
...
            elif isinstance(value, list):
                for exchange in value:
                    if isinstance(exchange, imod.mf6.GWFGWF):
                        exchange.write(
                            exchange.packagename(), globaltimes, write_context
                        )
```

We require something similar in the ``dump`` method.

luitjansl:
Fixes #679

# Description
expands the dump and from_file methods of Simulation to be able to deal with exchange files.
Tests were added. 
A follow-up issue may be needed, as gwfgwt exchanges are generated in the write-method and so in some cases the write method must be called first, before the dump method. 

# Checklist

- [X] Links to correct issue
- [X] Unit tests were added

### DIFF
--- a/imod/mf6/simulation.py
+++ b/imod/mf6/simulation.py
@@ -815,7 +815,7 @@ class Modflow6Simulation(collections.UserDict):
                     directory
                 ).as_posix()
             elif key in ["gwtgwf_exchanges","split_exchanges"]:
-                toml_content[key] = defaultdict(list)
+                toml_content[key] = collections.defaultdict(list)
                 for exchange_package in self[key]:
                     exchange_type, filename, _, _ = exchange_package.get_specification()
                     exchange_class_short = type(exchange_package).__name__
--- a/imod/mf6/simulation.py
+++ b/imod/mf6/simulation.py
@@ -814,7 +814,7 @@ class Modflow6Simulation(collections.UserDict):
                 toml_content[cls_name][key] = model_toml_path.relative_to(
                     directory
                 ).as_posix()
-            elif key == "split_exchanges":
+            elif key in ["gwtgwf_exchanges","split_exchanges"]:
                 toml_content[key] = {}
                 for exchange_package in self[key]:
                     exchange_type, filename, _, _ = exchange_package.get_specification()
@@ -857,7 +857,7 @@ class Modflow6Simulation(collections.UserDict):
         simulation = Modflow6Simulation(name=toml_path.stem)
         for key, entry in toml_content.items():
            
-            if key != "split_exchanges":
+            if not key in ["gwtgwf_exchanges", "split_exchanges"]:
                 item_cls = classes[key]
                 for name, filename in entry.items():
                     path = toml_path.parent / filename
--- a/imod/mf6/simulation.py
+++ b/imod/mf6/simulation.py
@@ -814,6 +814,17 @@ class Modflow6Simulation(collections.UserDict):
                 toml_content[cls_name][key] = model_toml_path.relative_to(
                     directory
                 ).as_posix()
+            elif key == "split_exchanges":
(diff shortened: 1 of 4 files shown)


imod/mf6/simulation.py
```diff
@@ -814,6 +814,17 @@ def dump(
                 toml_content[cls_name][key] = model_toml_path.relative_to(
                     directory
                 ).as_posix()
+            elif key in ["gwtgwf_exchanges","split_exchanges"]:
```
Manangka, line 817: We are seeing a lot of ~~if else statements~~ customn logic being added due to the exchange lists. It would be good to see if a refactor could simplify stuff. For instance we could add an ExchangeManager or ExchangeCollection object to hold the exchanges. This logic can then be moved to a dump method in that new class

Manangka (changes requested):

imod/mf6/simulation.py
luitjansl: that's not a bad idea, but the exchanges are just packages, so then I would prefer to make a package list/collection class for that . Since a `Model` is also a list of packages with some extra features, I think that if we make  a `PackageList`  then `Model ` should inherit from it (and use the same dump and from_file methods). But a refactoring like that deserves its own issue.

Manangka: That is an interesting idea. I will create a separate issue for it and will approve this one

Manangka (approved):

imod/mf6/simulation.py
```diff
@@ -814,6 +814,17 @@ def dump(
...
+                toml_content[key] = {}
+                for exchange_package in self[key]:
+                    exchange_type, filename, _, _ = exchange_package.get_specification()
+                    exchange_class_short = type(exchange_package).__name__
+                    if not exchange_class_short in toml_content[key].keys():
+                         toml_content[key][exchange_class_short]=[]
+                    path = f"{filename}.nc"
+                    exchange_package.dataset.to_netcdf(directory / path)
+                    toml_content[key][exchange_class_short].append(path)
```
JoerivanEngelen, line 826: You can reduce this code with a ``defaultdict`` in the native ``collections`` module:

```suggestion
                toml_content[key] = defaultdict(list)
                for exchange_package in self[key]:
                    exchange_type, filename, _, _ = exchange_package.get_specification()
                    exchange_class_short = type(exchange_package).__name__
                    path = f"{filename}.nc"
                    exchange_package.dataset.to_netcdf(directory / path)
                    toml_content[key][exchange_class_short].append(path)
```

JoerivanEngelen (commented):
Looks good, I have one suggestion to reduce the code a bit

PR #850, opened 2024-02-16
Issue #772 set repeat stress

### LINKED ISSUE #772: ``set_repeat_stress`` method got removed without warning
We are getting reports that users who were using the ``set_repeat_stress`` method and can't access it anymore after the 0.15.0 release.
This change passed by unnoticed to the development team, as the distinction between public and private API is not yet clearly communicated.
``set_repeat_stress`` method was public API and commonly applied, so at minimum some DeprecationWarnings needed to be thrown during a grace period if we wanted to remove it.

Before users had to do:

```python
riv = imod.mf6.River(...)
riv.set_repeat_stress(repeat_periods)
```

Users now have to go through this:

```python
from imod.mf6.utilities.package_utils import get_repeat_stress

repeat_stress = get_repeat_stress(repeat_periods) # args before provided to River.set_repeat_stress
imod.mf6.River(..., repeat_stress=repeat_stress)
```

This is uncomfortable API as the user needs to find an obscure utility function stowed away deep in some module.

Things to do at minimum:

- [ ] Add method removal to the changelog with some instructions how scripts should be updated
- [ ] EITHER: reintroduce ``set_repeat_stress`` for now until further decision has been made 
- [ ] OR: decide and implement what should be appropriate API to assign a ``repeat_stress`` variable with similar ease as before. This could be with a method (mutating the object, which is not what we really) or a utility function at a reachable location. This requires fixing #622 first

Refinement
-------------

- Until #622 is fixed, we decided to reintroduce the ``set_repeat_stress`` method for a grace period.
- The ``repeat_stress`` argument in the constructor and the changes it makes should be preserved
- The ``set_repeat_stress`` method can use all the backend of how stress periods are repeated via the constructor.
- Throw FutureDeprecationWarning mentioning when calling ``set_repeat_stress`` that this will be changed somewhere in the Future.

luitjansl:
Fixes #772

# Description
reinstated a method (set_period_data) that was removed, but which was used in user scripts. It is now back, with a deprecation warning. 

# Checklist

- [X] Links to correct issue
- [X] Unit tests were added

### DIFF
--- a/imod/mf6/boundary_condition.py
+++ b/imod/mf6/boundary_condition.py
@@ -14,7 +14,8 @@ from imod.mf6.auxiliary_variables import (
 from imod.mf6.package import Package
 from imod.mf6.write_context import WriteContext
 from imod.typing.grid import GridDataArray
-
+from imod.wq.timeutil import to_datetime
+import warnings
 
 def _dis_recarr(arrdict, layer, notnull):
     # Define the numpy structured array dtype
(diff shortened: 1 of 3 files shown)


imod/mf6/boundary_condition.py
```diff
@@ -77,6 +79,27 @@ def __init__(self, allargs: Mapping[str, GridDataArray | float | int | bool | st
...
+        Parameters
+        ----------
+        times: Dict of datetime-like to datetime-like.
+            The data of the value datetime is used for the key datetime.
+        """
+        warnings.warn(
+            f"""{self.__class__.__name__}.set_repeat_stress(...) is deprecated.
+            In the future, add repeat stresses as constructor parameters. 
+            An object containing them can be created using 'get_repeat_stress', as follows:
+            repeat_stress = get_repeat_stress(repeat_periods) # args before provided to River.set_repeat_stress
```
JoerivanEngelen, line 95: The main user unfriendly thing about ``get_repeat_stress`` is its location: stowed away in an obscure module. Add its location:

``from imod.mf6.utilities.package_utils import get_repeat_stress``

Furthermore:
We are still unsure about this utility location. So please also add the location of this utility might change in the future.

JoerivanEngelen (approved):
One comment, approving in advance.

imod/mf6/boundary_condition.py
luitjansl: done

PR #852, opened 2024-02-16
Issue #838 api package

### LINKED ISSUE #838: Add support for API-package of MODFLOW 6
The API package was created to provide a convenient way to touch MODFLOW 6 internals, outside the package levels. We need this package for the coupling between MODFLOW 6 - Ribasim - MetaSWAP. The package only contains an options and dimensions block. The actual data and nodemapping will be set via the imod_coupler. The minimum input than looks like:

 begin options
 end options
  
 begin dimensions
   MAXBOUND           1
 end dimensions

luitjansl:
Fixes #838

# Description
adds an api package. Its presence allows one to use the xmipy  api in combination with libmf6.dll.

# Checklist

- [X] Links to correct issue
- [X] Unit tests were added


imod/mf6/api_package.py
```diff
@@ -0,0 +1,50 @@
...
+from imod.schemata import (
+    DTypeSchema,
+)
+
+_init_schemata = {
+    "maxbound": [DTypeSchema(np.integer)],
+}
+
+class ApiPackage(Package):
+    """
```
JoerivanEngelen, line 13: The docstring still lacks a general explanation what the package does. I think your message in the changelog is a good start: 

"interact with libMF6.dll through its API."

Could you write that to a full sentence for the general explanation?

imod/mf6/api_package.py
```diff
@@ -0,0 +1,50 @@
...
+        the listing file for every stress period time step in which "BUDGET
+        PRINT" is specified in Output Control. If there is no Output Control
+        option and PRINT FLOWS is specified, then flow rates are printed for the
+        last time step of each stress period.
+        Default is False.
+    save_flows: ({True, False}, optional)
+        Indicates that constant head flow terms will be written to the file
+        specified with "BUDGET FILEOUT" in Output Control. Default is False.
+    
+    Note: this package can be added to both flow and transport models.
```
JoerivanEngelen, line 33: Sphinx has a special directive for notes, which makes them render nicer.

```suggestion
    .. note::
       This package can be added to both flow and transport models.
```

See: https://sublime-and-sphinx-guide.readthedocs.io/en/latest/notes_warnings.html

imod/mf6/api_package.py
luitjansl: done

imod/mf6/package.py
```diff
@@ -114,6 +119,8 @@ def _initialize_template(pkg_id):
             fname = f"gwt-{pkg_id}.j2"
         elif pkg_id in EXCHANGE_PACKAGES:
             fname = f"exg-{pkg_id}.j2"
+        elif pkg_id in API_PACKAGES:
+            fname = f"api-{pkg_id}.j2"
```
Manangka, line 123: I'm not sure we need a separate API_PACKAGES group. We are not adding more api packages are we?
Further the output file will be named api-api which is also a bit strange.
We could handle the Api package in a similair way like the ims and tdis package as it is one of a kind

Manangka (changes requested):

imod/mf6/package.py
luitjansl: done

Manangka (approved):

PR #864, opened 2024-02-22
Issue #578 now can write after splitting

### LINKED ISSUE #578: after splitting a simulation, the simulation cannot be written to file anymore
In GitLab by @luitjansl on Oct 12, 2023, 10:41

if we have a simulation object and call "split" on it, then the split function returns a new simulation. The original simulation should still be usable, but calling "write" on it produces an error. In pseudocode:

simulation_A = create_some_simulation()

simulation_B = simulation_A.split(label_array)

simulation_A.write(...) #produces error

Note:
It happens also without splitting. Calling the write method twice causes the same issue

Refinement
-------------
- debug until source of the error has been found
- possible mutations to original object: 
    - GWF-GWT exchange packages are added during the write method.
    - Any other mutations done during writing?
    - Any other mutations done during splitting?
- Add test:
    - Copy simulation A
    - Split simulation A
    - Simulation A should be equal to its copy

luitjansl:
Fixes #578 

# Description
The simulation could not be written to file after splitting, because  some  packages of the split simulation were 
the same object as those packages of the source simulation, so that any further modifications to the packages were also applied to the source simulation, making it inconsistent. 
This has been fixed.
Another comment in the Issue stated that calling "write" on a simulation twice also gave an issue.
This could not be reproduced, but a test was added just in case

# Checklist
- [X] Links to correct issue
- [X] Unit tests were added

### DIFF
--- a/imod/mf6/simulation.py
+++ b/imod/mf6/simulation.py
@@ -5,6 +5,7 @@ import copy
 import pathlib
 import subprocess
 import warnings
+from copy import deepcopy
 from pathlib import Path
 from typing import Any, Callable, DefaultDict, Iterable, Optional, Union, cast
 
@@ -15,7 +16,6 @@ import tomli
 import tomli_w
 import xarray as xr
 import xugrid as xu
-from copy import deepcopy
 
 import imod
 import imod.logging
--- a/imod/mf6/simulation.py
+++ b/imod/mf6/simulation.py
@@ -15,6 +15,7 @@ import tomli
 import tomli_w
 import xarray as xr
 import xugrid as xu
+from copy import deepcopy
 
 import imod
 import imod.logging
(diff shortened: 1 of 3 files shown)


Manangka (approved):

JoerivanEngelen (approved):
Looks good, 2 things:

* could you add ``Issue #864`` to the issue title?
* could you add this fix to the changelog as well?

imod/mf6/simulation.py
```diff
@@ -999,7 +1000,7 @@ def split(self, submodel_labels: GridDataArray) -> Modflow6Simulation:
 
         new_simulation = imod.mf6.Modflow6Simulation(f"{self.name}_partioned")
         for package_name, package in {**original_packages}.items():
-            new_simulation[package_name] = package
```
JoerivanEngelen, line 1002: Might be interesting to add a ``.copy`` method to the Package object.

luitjansl: Sure,. I'm used to using copy.deepcopy for everything but  for some users it may be helpful.

PR #866, opened 2024-02-22
Issue #601 deepcopy

### LINKED ISSUE #601: copy.deepcopy does not work on simulation or model class
In GitLab by @luitjansl on Oct 24, 2023, 14:36

To reproduce, take an example ( eg Henry) and add "`import copy`" to the import statements. Then add `copied_simulation = copy.deepcopy(simulation)`

in it, after the model has been completely assembled, but before it is being written or run ( at the time of writing of this issue, line 244).This results in an error when executing that line:

Template.\__new_\_() missing 1 required positional argument: 'source'

File "D:\\dev\\imod-python\\examples\\mf6\\Henry.py", line 244, in csim = copy.deepcopy(simulation) TypeError: Template.\__new_\_() missing 1 required positional argument: 'source'

Further experimenting revealed that an error is produced also when deepcopying a model, but not a package

luitjansl:
Fixes #601

# Description
Fixes deepcopying for model and simulation objects. 
The issue was that the jinja-template was  an instance variable in these classes, and when making a deepcopy python attempted to copy them. But the jinja templates are not copyable, so this caused a crash. Now the templates are class variables for model and simulation, so when making a deepcopy, the same template object is used for   the copy as for the source. 

# Checklist

- [X] Links to correct issue
- [X] Unit tests were added

### DIFF
--- a/imod/mf6/model.py
+++ b/imod/mf6/model.py
@@ -27,8 +27,6 @@ from imod.schemata import ValidationError
 from imod.typing import GridDataArray
 
 
-
-
 class Modflow6Model(collections.UserDict, abc.ABC):
     _mandatory_packages: Tuple[str, ...] = ()
     _model_id: Optional[str] = None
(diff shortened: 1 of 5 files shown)


JoerivanEngelen (approved):
Looks good, thanks for fixing this!

PR #867, opened 2024-02-23
Issue #921 Update uzf

### LINKED ISSUE #921: Support reading multiple fluxes for one package
For advanced stress packages (e.g. UZF). For example, for a package UnsaturatedZonePackage with the name "unsat", there are multiple fluxes which are saved, so just picking one and using its name "unsat" is insufficient.

HendrikKok:
Fixes #921

# Description
This branch contains all fixes that I needed to work with the UZF-package. The branch contains:

1- Changes to UZF- package methods
2- Changes to the open_cbc methods to deal with cbc-files of advanced packages
3- Changes to the UZF-tests + all tests that deal with budget output

There are some breaking changes in budget output names:
1- The new names support now support output of multiple equal boundary packages, since the package name is used.
2- The budget arrays now contain nans if a package is not active, since zero is a valid budget output.

# Checklist

- [x] Links to correct issue
- [x] Unit tests were added


imod/mf6/out/cbc.py
```diff
@@ -288,8 +297,9 @@ def read_imeth6_budgets_dense(
     Three-dimensional array of floats
     """
     # Allocates a dense array for the entire domain
-    out = np.zeros(size, dtype=np.float64)
+    out = np.full(size, np.nan, dtype=np.float64)
```
Huite, line 300: Using NaN instead of zeros is an okay choice, but it should be mentioned in the changelog.

I think I used zeros here because iMODFLOW, iMOD-WQ do the same.
It makes adding different water balance terms a bit easier. Maybe open a separate issue for this?

imod/mf6/out/__init__.py
```diff
@@ -227,6 +227,7 @@ def open_cbc(
     flowja: bool = False,
     simulation_start_time: Optional[np.datetime64] = None,
     time_unit: Optional[str] = "d",
+    uzf: bool = False,
```
Huite, line 230: I did a quick search across the mf6 IO docs, there are multiple packages with their own budget file.

In this case, I think I'd prefer having a specific function to deal with the cbc file instead? `open_cbc_uzf` or something? Because we might need to add additional methods for the water mover, the lake package, etc; and we don't want to have five boolean args.

imod/mf6/uzf.py
```diff
@@ -350,7 +358,7 @@ def _determine_vertical_connection(self, uzf_number):
         return uzf_number.shift(layer=-1, fill_value=0)
 
     def _package_data_to_sparse(self):
-        notnull = self.dataset["landflag"].values == 1
+        notnull = np.isfinite(self.dataset["landflag"].values)
```
Huite, line 361: If this is the better condition, maybe just use `self.dataset["landflag"].notnull().to_numpy()`

imod/mf6/out/__init__.py
HendrikKok: I removed this optional argument. I changed the output budget keys to a more general form that should work for all budget files. To identify the advanced packages, I check the header keys for the label 'gwf'. As far as I could see, that's unique to the advance packages.

imod/mf6/uzf.py
HendrikKok: ok

HendrikKok:
This branch contains all fixes that I needed to work with the UZF-package. The branch contains:

1-	Changes to UZF- package methods
2-	Changes to the open_cbc methods to deal with cbc-files of advanced packages
3-	Changes to the UZF-tests + all tests that deal with budget output

There are some breaking changes in budget output names:
1-	The new names support now support output of multiple equal boundary packages, since the package name is used.
2-	The budget arrays now contain nans if a package is not active, since zero is a valid budget output.

imod/mf6/out/common.py
```diff
@@ -19,3 +19,13 @@ def _to_nan(a: FloatArray, dry_nan: bool) -> FloatArray:
...
+
+
+def get_header_advanced_package(
+    headers: Dict[str, List[Any]],
+) -> Any:
+    for key, header in headers.items():
+        # multimodels have a gwf-gwf budget for flow-ja-face between domains
+        if not "flow-ja-face" in key and "gwf_" in key:
+            return header[0]
+    return None
```
JoerivanEngelen, line 31: I don't really get what this function is doing:

Why does it return ``header[0]`` and not ``header``? Does this mean the function name is incorrect, or are we actually returning the first header in the list? If the latter:

```suggestion
def get_header_advanced_package(
    headers: Dict[str, List[Any]],
) -> Any:
    for key, header_ls in headers.items():
        # multimodels have a gwf-gwf budget for flow-ja-face between domains
        if not "flow-ja-face" in key and "gwf_" in key:
            return header_ls[0]
    return None
```

In that case, maybe better function name is: ``get_first_header_advanced_package``

UPDATE: I see in other part of the codebase:
``for key, header_list in headers.items():``
So I think that is clearer here.

imod/mf6/out/dis.py
```diff
@@ -210,6 +211,8 @@ def open_imeth6_budgets(
     cbc_path: str, pathlib.Path
     grb_content: dict
     header_list: List[Imeth1Header]
+    return_variable: str
+    return_id: np.ndarray | None
```
JoerivanEngelen, line 215: See earlier comment: rename ``return_id`` to ``indices``

imod/mf6/simulation.py
```diff
@@ -638,7 +638,7 @@ def _merge_and_assign_exchange_budgets(self, cbc: GridDataset) -> GridDataset:
         exchange_budgets = cbc[exchange_names].to_array().sum(dim="variable")
         cbc = cbc.drop_vars(exchange_names)
         # "gwf-gwf" or "gwt-gwt"
-        exchange_key = exchange_names[0].split("_")[0]
+        exchange_key = exchange_names[0].split("_")[1]
```
JoerivanEngelen, line 641: Why was this necessary? It changes behavior for gwf-gwt exchanges, was that required?

imod/mf6/uzf.py
```diff
@@ -182,12 +187,12 @@ class UnsaturatedZoneFlow(AdvancedBoundaryCondition):
         "theta_sat": [IdentityNoDataSchema("kv_sat"), AllValueSchema(">=", 0.0)],
         "theta_init": [IdentityNoDataSchema("kv_sat"), AllValueSchema(">=", 0.0)],
         "epsilon": [IdentityNoDataSchema("kv_sat")],
-        "infiltration_rate": [IdentityNoDataSchema("kv_sat")],
-        "et_pot": [IdentityNoDataSchema("kv_sat")],
-        "extinction_depth": [IdentityNoDataSchema("kv_sat")],
-        "extinction_theta": [IdentityNoDataSchema("kv_sat")],
-        "root_potential": [IdentityNoDataSchema("kv_sat")],
-        "root_activity": [IdentityNoDataSchema("kv_sat")],
+        "infiltration_rate": [IdentityNoDataSchema("stress_period_active")],
```
JoerivanEngelen, line 190: Could you add a comment why ``"stress_period_active"`` is required, and not ``"kv_sat"`` for these variables? Why is ``"kv_sat"`` still used for the ``IdentityNoDataSchema`` for ``"epsilon"``?

imod/mf6/uzf.py
```diff
@@ -341,16 +352,19 @@ def _check_options(
 
     def _create_uzf_numbers(self, landflag):
         """Create unique UZF ID's. Inactive cells equal 0"""
-        return np.cumsum(np.ravel(landflag)).reshape(landflag.shape) * landflag
+        active_nodes = xr.ones_like(landflag).where(landflag.notnull(), other =0.0)
```
JoerivanEngelen, line 355: I think this can be shortened to:

```suggestion
        active_nodes = landflag.notnull().astype(np.int8)
```

imod/mf6/uzf.py
```diff
@@ -341,16 +352,19 @@ def _check_options(
...
+        return np.nancumsum(active_nodes).reshape(landflag.shape) * active_nodes
 
     def _determine_landflag(self, kv_sat):
-        return (np.isfinite(kv_sat)).astype(np.int32)
+        """ returns the landflag for uzf-model. Landflag == 1 for top active UZF-nodes """
```
JoerivanEngelen, line 359: Thanks for adding a docstring!

imod/mf6/uzf.py
```diff
@@ -341,16 +352,19 @@ def _check_options(
...
+        land_nodes = xr.ones_like(kv_sat,dtype = np.int32).where(kv_sat.layer == get_upper_active_layer_number(kv_sat),other = 0)
```
JoerivanEngelen, line 360: [This can be shortened further with this utility function](https://deltares.github.io/imod-python/api/generated/prepare/imod.prepare.get_upper_active_grid_cells.html)

```suggestion
        land_nodes = get_upper_active_grid_cells(kv_sat).astype(np.int32)
```

JoerivanEngelen (commented):
I think the PR makes some necessary changes to iMOD Python, but I have some comments. 

Furthermore, could you update the changelog with all your changes? Something like:

```
Added
-------
- Support reading fluxes in the CBC file for boundary packages that contain multiple fluxes (e.g. UZF)

Changed
----------
- Package type added to name of CBC output in fluxnames. E.g. a River package named ``primary-sys`` will get fluxname ``riv_primary-sys``.
- Fluxes in inactive cells now have value ``np.nan`` instead of 0.0
```

UPDATE: I also miss a test where the UZF cbc fluxes are read. I think that can be added to the test in ``imod/tests/test_mf6/test_mf6_uzf_model.py``?

imod/mf6/uzf.py
HendrikKok: done

imod/mf6/uzf.py
HendrikKok: done

imod/mf6/simulation.py
HendrikKok: Yes, since we have an extra '-' in de variable names.

JoerivanEngelen (approved):
Looks good, I got one minor comment

PR #880, opened 2024-02-28
Issue #724 Validation package settings

### LINKED ISSUE #724: Add validation for incompatible settings
Some settings are incompatible with each other, for example:

- NPF: ``alternative_cell_averaging`` is not compatible ``XT3D``
- NPF: ``RHS`` can only be used when ``XT3D`` is turned on
- IMS: ``NO_PTC`` only useful when Model option ``newton`` is used
- DSP: ``XT3D_RHS`` only used when NPFs ``XT3D`` is turned on

We need to create some ValidationSchemes for this, that can be used with other packages.

JoerivanEngelen:
Fixes #724

# Description
- Adds new validation scheme to compare setting incompatibility
- Implements validation of incompatible settings in the NPF and DSP package.
- Validation of IMS and model settings (newton) would require quite some extra code for little gains, so I scrapped that.

# Checklist
- [x] Links to correct issue
- [x] Unit tests were added

### DIFF
--- a/imod/tests/test_mf6/test_mf6_dsp.py
+++ b/imod/tests/test_mf6/test_mf6_dsp.py
@@ -1,10 +1,13 @@
 import pathlib
+import re
 import textwrap
 
 import numpy as np
+import pytest
 import xarray as xr
 
 import imod
+from imod.schemata import ValidationError
 
 
 def test_dispersion_default():
@@ -34,13 +37,12 @@ def test_dispersion_default():
 def test_dispersion_options():
     directory = pathlib.Path("mymodel")
     globaltimes = [np.datetime64("2000-01-01")]
-    disp = imod.mf6.Dispersion(1e-4, 1.0, 10.0, 1.0, 2.0, 3.0, True, True)
+    disp = imod.mf6.Dispersion(1e-4, 1.0, 10.0, 1.0, 2.0, 3.0, True, False)
     actual = disp.render(directory, "dsp", globaltimes, True)
     expected = textwrap.dedent(
         """\
       begin options
         xt3d_off
-        xt3d_rhs
       end options
 
       begin griddata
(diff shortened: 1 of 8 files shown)


imod/tests/test_mf6/test_mf6_dsp.py
```diff
@@ -61,6 +90,15 @@ def test_dispersion_options():
 
     assert actual == expected
 
+def test_dispersion_erronous_options():
+    message = textwrap.dedent(
```
luitjansl, line 94: I wouldn't test on the message. In most pytest.raises tests this is not done, and it will cause a test to fail if the message is edited for some reason.

imod/typing/__init__.py
```diff
@@ -9,5 +9,6 @@
 
 GridDataArray: TypeAlias = Union[xr.DataArray, xu.UgridDataArray]
 GridDataset: TypeAlias = Union[xr.Dataset, xu.UgridDataset]
+ScalarDataArray: TypeAlias = Union[xr.DataArray, xu.UgridDataArray]
```
luitjansl, line 12: This is the same type as GridDataArray.
I would prefer ScalarAsDataArray to avoid confusion with dataarrays that have scalars for array elements.
You could also create a derived type of xr.DataArray which has a single element, and use that derived type in the TypeAlias.   
Is the UgriDataArray  option really necessary? This type seems to be used here for options like "xt3d_off" which are not really structured or unstructured.

luitjansl (commented):
some comments. The issue also mentions IMS: NO_PTC only useful when Model option newton is used.
Please edit the issue to remove that, or create a follow-up issue for it.

imod/tests/test_mf6/test_mf6_dsp.py
luitjansl: maybe we should go for numbered error messages then. Then we can check on the error number rather than on the precise string. That's more an idea for the future and would not be part of this PR.

imod/typing/__init__.py
JoerivanEngelen: I renamed it to ScalarAsDataArray. UgridDataArray option is necessary, to as object types are preserved, for example when UgridDataset[scalar_variable] -> UgridDataArray

luitjansl (approved):
looks good

Issue #683, opened 2023-12-07
Partioned CBC output cannot be merged when boundary condition not present in every partition


Manangka:
In GitLab by @JoerivanEngelen on Dec 7, 2023, 15:09

When one boundary condition is missing in one partition, it is dropped from the model (logically). This causes certain variables (e.g. ```cbc["wel"]```) to be missing in a certain partition. When merging, this causes xugrid to throw a clear error: 

```
ValueError: These variables are present in some partitions, but not in others: {'well'}
```

Manangka:
In GitLab by @JoerivanEngelen on Dec 7, 2023, 15:10

marked this issue as related to #658

luitjansl:
this issue was solved along with #846  in https://github.com/Deltares/imod-python/pull/881

PR #881, opened 2024-02-28
Issue #846 unstructured concentric

### LINKED ISSUE #846: add  ( and debug) concentric partition to tests on the circle model
The circle model is used in different tests for unstructured grids.  In test file test_mf6_partitioning_unstructured.py
we wanted to add a new case, which partitions the circle in a concentric fashion, with one partition at the model boundary and an internal partition that has no boundaries ( except recharge).  But almost all test fail that use this partition, so it was not added, and we should investigate why they fail. The new case is shown below:

`class PartitionArrayCases:`
 ... ( 2 existing cases, and then:)
```
    def case_concentric(self, idomain_top) -> xu.UgridDataArray:
        centroids = idomain_top.ugrid.grid.centroids
        dist = np.sqrt( centroids[:,0]* centroids[:,0] +  centroids[:,1]* centroids[:,1])
        concentric = zeros_like(idomain_top)
        concentric.values =  np.where(dist < 500, 0, 1)
        return  concentric 
```

luitjansl:
Fixes #846  and #683

# Description
adds a concentric partition to the unstructured partitioning tests. Cleanup of the test file because various new functions ( simulation.open_heads, model.mask_all_packages etc) are now available. Introduces balance comparison to the test- although some require high tolerances. 
Fixed an issue ( #683) which made it unable to merge balance results if not all boundary conditions were present in all partitions.

# Checklist

- [X] Links to correct issue
- [ ] Unit tests were added

### DIFF
--- a/imod/tests/test_mf6/test_multimodel/test_mf6_partitioning_unstructured.py
+++ b/imod/tests/test_mf6/test_multimodel/test_mf6_partitioning_unstructured.py
@@ -32,6 +32,12 @@ class PartitionArrayCases:
         three_parts.values[144:] = 2
         return three_parts
 
+    def case_concentric(self, idomain_top) -> xu.UgridDataArray:
+        centroids = idomain_top.ugrid.grid.centroids
+        dist = np.sqrt( centroids[:,0]* centroids[:,0] +  centroids[:,1]* centroids[:,1])
+        concentric = zeros_like(idomain_top)
+        concentric.values =  np.where(dist < 500, 0, 1)
+        return  concentric 
 
 class WellCases:
     def case_one_well(self):
@@ -131,6 +137,7 @@ class HorizontalFlowBarrierCases:
 def test_partitioning_unstructured(
     tmp_path: Path, circle_model: Modflow6Simulation, partition_array: xu.UgridDataArray
 ):
+    # %%
     simulation = circle_model
     # Increase the recharge to make the head gradient more pronounced.
     simulation["GWF_1"]["rch"]["rate"] *= 100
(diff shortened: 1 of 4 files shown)


imod/tests/test_mf6/test_multimodel/test_mf6_partitioning_unstructured.py
```diff
@@ -435,20 +360,22 @@ def test_partition_transport_multispecies(    tmp_path: Path,
...
     conc_new = new_circle_model.open_concentration()
     head_new = new_circle_model.open_head()
+    head_new = head_new.ugrid.reindex_like(head)   
+    conc_new = conc_new.ugrid.reindex_like(conc)
+
     
-    np.testing.assert_allclose(conc.values, conc_new["concentration"].values, rtol=1e-5, atol=1e-3)
-    np.testing.assert_allclose(head.values, head_new["head"].values, rtol=1e-5, atol=1e-3)
+    np.testing.assert_allclose(conc.values, conc_new["concentration"].values, rtol=4e-4, atol=5e-3)
+    np.testing.assert_allclose(head.values, head_new["head"].values, rtol=0.008, atol=0.15)
```
JoerivanEngelen, line 378: I see the criteria are set at seemingly scary levels, however heads in the example range from 2000 to 0 m. I think it is good to add a comment that heads range from 2000 to 0 m (implying that we shouldn't worry about too much about it...).

JoerivanEngelen (approved):
Looks good, one comment. Approving in advance

imod/tests/test_mf6/test_multimodel/test_mf6_partitioning_unstructured.py
luitjansl: done. I now set the higher tolerances only for the concentric partition scheme, and added a comment.

PR #882, opened 2024-02-28
Issue #622 cleanup utilities part 1.

### LINKED ISSUE #622: Cleanup utilities part 1

JoerivanEngelen:
Fixes #622

# Description

This PR separates the ``imod.util`` into separate modules.
Care has been taken that the migration doesn't affect backwards compatibility with public API utility functions.
Furthermore, ``imod.util.round_extent`` has been made available again for backwards compatibility, but now throws a DeprecationWarning if used.

# Checklist

- [x] Links to correct issue
- [x] Unit tests were added


luitjansl (approved):
Looks good. But to what extent will this impact user scripts? No deprecation warnings are given as far as I can see. Do we want that? I guess for most users it will be fine as they are unlikely to use these methods directly (but do we know that?)

Huite:
I'd say `cd`, `temporary_directory`, `ignore_warnings`, `ugrid2d_data`, `mdal_compliant_ugrid2d`, `from_mdal_compliant`, `to_ugrid2d`, `empty_2d`, `empty_2d_transient`, `empty_3d`, `empty_3d_transient`, `where`, `replace`, `values_within_range` should remain available under `imod.util`.

Just keep `imod.util` and import all the methods mentioned here.

Furthermore, from I would also include `round_extent` from `imod.prepare.spatial`, because I've had to copy it to the scripts of multiple people.

JoerivanEngelen:
@luitjansl 
See my reply to Huite's comments for an answer to your questions. I only opted for one DeprecationWarning where a function was moved to the ``prepare`` module ages ago (warning should have been there in the first place).

@Huite 

> I'd say `cd`, `temporary_directory`, `ignore_warnings`, `ugrid2d_data`, `mdal_compliant_ugrid2d`, `from_mdal_compliant`, `to_ugrid2d`, `empty_2d`, `empty_2d_transient`, `empty_3d`, `empty_3d_transient`, `where`, `replace`, `values_within_range` should remain available under `imod.util`.

Thanks for the list, I made sure all these are available under the imod.util namespace.
 
> Just keep `imod.util` and import all the methods mentioned here.

 I've made sure users can still import everything as they were used to:
``from imod.util import where``
by importing these functions in the imod/util/__init__.py file.
I also added a new unit test to test that this works.

> Furthermore, from I would also include `round_extent` from `imod.prepare.spatial`, because I've had to copy it to the scripts of multiple people.

Ah, I didn't know about that one, made sure it is imported and a DeprecationWarning is thrown when users do this.

PR #891, opened 2024-02-29
Issue #818 verify one flowmodel

### LINKED ISSUE #818: throw error when splitting a simulation with different models that use different discretizations
Since different models in a simulation can use different discretizations, we should check that all the models that we are going to split in a simulation are using the same discretizations. If this is not the case throw an error. 

Refinement
--------------
- When calling ``.split`` method, check that only one flow model is assigned to simulation. 
- Multiple transport models can exist for each species.
- Add test
- Add this validation to simulation.regrid_like (as we do not support regridding multiple models with a separate grid). The workflow is always: first regrid, then split. Explicitly state this in error message.
- For masking and clipping, idem.

luitjansl:
Fixes #818

# Description
Throws an error if multiple flow models are present when clipping, regridding or splitting a simulation. 

# Checklist

- [X] Links to correct issue
- [X] Unit tests were added

### DIFF
--- a/docs/api/changelog.rst
+++ b/docs/api/changelog.rst
@@ -6,6 +6,11 @@ All notable changes to this project will be documented in this file.
 The format is based on `Keep a Changelog`_, and this project adheres to
 `Semantic Versioning`_.
 
+[Feature branch specific]
+-------------------------
+- Checks that only one flow model is present in a simulation when calling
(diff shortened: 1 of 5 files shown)


JoerivanEngelen (approved):
Looks good

PR #913, opened 2024-03-11
Issue #889 mask simulation

### LINKED ISSUE #889: Add mask method to Modflow6Simulation level
The mask method is still missing on simulation level, this can be easily implemented in the case of a single flow model with coupled transport models. With partitioned simulations, providing a simple api is more complicated. 

Easiest for now is to let users only mask simulations which are not split yet. If the simulation is split, users have to manually regrid individual models themselves.

Requirements
----------------
- If simulation split, throw error, guiding the user to mask individual model objects manually
- Be able to mask a single flow model with multiple transport models.

luitjansl:
Fixes #889 

# Description
Adds a masking method to the simulation object. It masks all the models that hang under the simulation, provided the simulation has not been split, and all the models have the same domain.

# Checklist

- [X] Links to correct issue
- [X] Unit tests were added


imod/typing/grid.py
```diff
@@ -279,4 +279,17 @@ def is_equal(array1: xr.DataArray, array2: xr.DataArray)->bool:
 
 @typedispatch
 def is_equal(array1: object, array2:object)->bool:
+    return False
+
+
+@typedispatch
+def is_same_domain(array1: xu.UgridDataArray, array2: xu.UgridDataArray)->bool:
```
Manangka, line 286: Instead of array1 and array2 i would call this grid1 and grid2

imod/mf6/simulation.py
```diff
@@ -1248,4 +1249,37 @@ def _update_buoyancy_packages(self)->None:
...
+        mask: GridDataArray,
+    ):
+        """
+        This function applies a mask to all models in a simulation, provided they use
+        the same discretization. 
+
+        Parameters
+        ----------
+        mask: xr.DataArray, xu.UgridDataArray of ints
+            idomain-like integer array. 1 sets cells to active, 0 sets cells to inactive,
```
Manangka, line 1266: This is basically the definition of the idomain. Are you masking the original domain or are you defining a new idomain? A method name like `alter_cell_passthrough` would also describe this method

Manangka (approved):

imod/typing/grid.py
```diff
@@ -279,4 +279,17 @@ def is_equal(array1: xr.DataArray, array2: xr.DataArray)->bool:
 
 @typedispatch
 def is_equal(array1: object, array2:object)->bool:
+    return False
+
+
+@typedispatch
+def is_same_domain(grid1: xu.UgridDataArray, grid2: xu.UgridDataArray)->bool:
+    return  grid1.coords.equals(grid2.coords) and grid1.ugrid.grid.equals(grid2.ugrid.grid)
```
JoerivanEngelen, line 287: Is the check on grids depending on the index order? Maybe we need a call to reindex_like?

JoerivanEngelen (commented):
I have some suggestions!

imod/mf6/simulation.py
```diff
@@ -1248,4 +1249,41 @@ def _update_buoyancy_packages(self)->None:
...
+        Masking will update the packages accordingly, blanking their input where needed,
+        and is therefore not a reversible operation.  
+
+        Parameters
+        ----------
+        mask: xr.DataArray, xu.UgridDataArray of ints
+            idomain-like integer array. 1 sets cells to active, 0 sets cells to inactive, 
+            -1 sets cells to vertical passthrough
+        """
+        if any([coord not in ["x", "y", "layer", "mesh2d_nFaces", "dx", "dy"] for coord in mask.coords]):
```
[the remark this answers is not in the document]
luitjansl: done (the typing.grid approach ). 
But the string "mesh2d_nFaces" is found  at numerous other locations in the code base. We may need some additional testing eg  with packages using all different face dimension names in a single simulation.

imod/typing/grid.py
luitjansl: indeed we may need a call to reindex_like, but I would leave that to the calling side.  I wouldn't expect a method like "is_same_domain" to modify the DataArray.

PR #922, opened 2024-03-18
Issue #807 split twice

### LINKED ISSUE #807: splitting a flow model twice leads to an error
the following test  fails: it splits the circle model once, but fails to do so a second time. In this example there is no transport. 
```
@pytest.mark.usefixtures("circle_model")
def test_slice_model_twice(tmp_path, circle_model):

    flow_model = circle_model["GWF_1"]
    active = flow_model.domain

    submodel_labels = zeros_like(active)
    submodel_labels = submodel_labels.drop_vars("layer")
    submodel_labels.values[ :, 50:] = 1
    submodel_labels = submodel_labels.sel(layer=0, drop=True)

    split_simulation_1 = circle_model.split(submodel_labels)
    print("split once")   
    split_simulation_2 = circle_model.split(submodel_labels)
    print("split twice!")    
```

luitjansl:
Fixes #807 

# Description
The issue itself was a bugreport, it said that after splitting a model, calling "split" a second time on the input model would result in a crash. This could not be reproduced. Still, a test was added to make sure we can split a model twice in the same way and that we get the same result in both cases.

# Checklist

- [X] Links to correct issue
- [X] Unit tests were added

### DIFF
--- a/imod/tests/test_mf6/test_multimodel/test_mf6_partitioning_unstructured.py
+++ b/imod/tests/test_mf6/test_multimodel/test_mf6_partitioning_unstructured.py
@@ -1,3 +1,4 @@
+from filecmp import dircmp
 from pathlib import Path
 
 import geopandas as gpd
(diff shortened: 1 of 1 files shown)


JoerivanEngelen (approved):
Looks good, thanks for fixing!

PR #925, opened 2024-03-19
Issue #924 Fix initialization of the ihc array in GWTGWT

### LINKED ISSUE #924: IHC array initialized incorrectly in GWTGWT
I'm running into an issue with the lhm-zz model, where the GWTGWT exchanges get incorrectly shaped arrays during writing. The GWFGWF exchange has no issues. The issue is fixed by changing this line in the initialization:

```python
            "ihc": xr.DataArray(np.ones_like(cl1, dtype=int)),
```

into:

```python
            "ihc": ones_like(cl1, dtype=int),
```

similar to the GWFGWF, then this is fixed. I think this is because the "ihc" array otherwise gets another dimensions "dim_0", instead of "index", and therefore doesn't get filtered properly in case inactive cells are located on multi-model borders.

JoerivanEngelen:
Fixes #924 

# Description
Fixes initialization of the ihc array in GWTGWT, concurrent with the GWFGWF exchanges.

# Checklist

- [x] Links to correct issue
- [ ] Unit tests were added

### DIFF
--- a/imod/mf6/gwtgwt.py
+++ b/imod/mf6/gwtgwt.py
@@ -9,6 +9,7 @@ from imod.mf6.auxiliary_variables import expand_transient_auxiliary_variables
 from imod.mf6.exchangebase import ExchangeBase
 from imod.mf6.package import Package
 from imod.typing import GridDataArray
+from imod.typing.grid import ones_like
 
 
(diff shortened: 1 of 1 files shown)


luitjansl (approved):
looks good

PR #933, opened 2024-03-25
Issue #429 regridding cache

### LINKED ISSUE #429: Refactor: Improve Regridder caching Modflow6 objects
In GitLab by @JoerivanEngelen on Jun 9, 2023, 17:33

Similar approach as in ``SingularTargetWeightsCache`` in ``disv_convserion.py`` module, but instead use Regridder types provided by package classes instead of an if else based on variable name.

Requirements:
- Cache on Simulation level instead of Package level
- Store regridding weights instead of regridder instances
- cache based on xy coordinates, this should be used to convert iMOD5 models as well, which can contain all kinds of perfidious grids.
   - The  ``SingularTargetWeightsCache`` approach can be followed: Hash x,y coordinates to get grid uniqueness
   - Structured grids can get their x, y coordinates hashed
   - Unstructured grids should also hash x,y coordinates (under ``ugrid.grid``)
- Ideally, the CachedRegridder, can operate in two contexts:
   - Regrid Modflow 6 model one grid to Modflow 6 model on other grid
   - During the (to be created) iMOD5 conversion: Regrid model data defined on different (structured) grids to one single grid

For testing, easiest way to test this is to create a modified fixture of the Simulation object, modify one or multiple packages to a different grid. If we turn the validation off, we should be able to regrid this to a single grid.

luitjansl:
Fixes #429

# Description
Now caches weight sets instead of regridders when regridding models. This saves re-computing weights, because the same weights are used for different regridders (if the regridder type is the same and only the regridding method is different). 
For example if we have 2 overlap regridders, but one is of method "mean" and the other one is of method "geometric_mean" they will use the same weights.

# Checklist

- [X] Links to correct issue
- [X] Unit tests were added


imod/typing/grid.py
```diff
@@ -306,3 +306,20 @@ def get_spatial_dimension_names( grid: xu.UgridDataArray)-> list[str]:
...
+def get_grid_geometry_hash( grid: xr.DataArray)-> int:
+    result = hash(pickle.dumps(grid["x"].values))
+    result += hash(pickle.dumps(grid["y"].values))
+    if " layer"  in grid.coords:
+        result += hash(pickle.dumps(grid["layer"].values))
+    return result
+
+@typedispatch
+def get_grid_geometry_hash( grid: xu.UgridDataArray)-> int:
+    g =  hash(pickle.dumps(grid.ugrid.grid))
```
Huite, line 320: For Ugrid2D and the type of regridding we're doing, it's not necessary to check the entire object; it will cause different hashes if some attrs are different.

I recommend checking only node_x, node_y, and face_node_connectivity. From the perspective of regridding of faces, these are the only ones that matter.

imod/typing/grid.py
```diff
@@ -306,3 +306,20 @@ def get_spatial_dimension_names( grid: xu.UgridDataArray)-> list[str]:
 @typedispatch
 def get_spatial_dimension_names( grid: object)-> list[str]:
     return []
+
+@typedispatch
+def get_grid_geometry_hash( grid: xr.DataArray)-> int:
+    result = hash(pickle.dumps(grid["x"].values))
```
Huite, line 312: It's worth researching/checking how fast the different options are for hashing numpy arrays. See for example: https://stackoverflow.com/questions/16589791/most-efficient-property-to-hash-for-numpy-array

I've thought about xxhash before, but I'm not sure it's worth it to add another dependency just for this use.

I don't think it's necessary to use cryptographic hashes from hashlib, but it's probably worth it to recheck what kind of algorithm `hash` is using behind the scenes.

imod/tests/test_mf6/test_mf6_regrid.py
```diff
@@ -115,9 +116,13 @@ def test_regrid_structured():
     """
     structured_grid_packages = create_package_instances(is_structured=True)
     new_grid = grid_data_structured(np.float64, 12, 2.5)
+    old_grid = grid_data_structured (np.float_, 1.0e-4, 5.0)
```
JoerivanEngelen, line 119: More consistent with declaration ``new_grid``:

```suggestion
    old_grid = grid_data_structured(np.float64, 1.0e-4, 5.0)
```

imod/typing/grid.py
```diff
@@ -306,3 +306,20 @@ def get_spatial_dimension_names( grid: xu.UgridDataArray)-> list[str]:
...
+    result += hash(pickle.dumps(grid["y"].values))
+    if " layer"  in grid.coords:
+        result += hash(pickle.dumps(grid["layer"].values))
```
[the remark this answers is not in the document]
JoerivanEngelen: Agreed, ``layer`` coord is irrelevant, since regridding in ``xugrid`` is purely 2D

examples/mf6/different_ways_to_regrid_models.py
```diff
@@ -94,8 +94,9 @@
 # in this example we'll regrid the npf package manually and the rest of the packages using default methods.
```
JoerivanEngelen, line 95: I think this is a nice opportunity to explain something about the ``RegridderWeightsCache``! Please add one or two sentences what it does and why it is useful.

imod/mf6/utilities/regrid.py
```diff
@@ -23,10 +23,10 @@
...
 from imod.schemata import ValidationError
-from imod.typing.grid import GridDataArray, ones_like
+from imod.typing.grid import GridDataArray, get_grid_geometry_hash, ones_like
 
 
-class RegridderInstancesCollection:
+class RegridderWeightsCache:
     """
     This class stores any number of regridders that can regrid a single source grid to a single target grid.
     By storing the regridders, we make sure the regridders can be re-used for different arrays on the same grid.
```
JoerivanEngelen, line 32:
```suggestion
    By storing the regridders, we make sure the regridders can be re-used for different arrays on the same grid.
    Regridders are stored based on their type (`see these docs<https://deltares.github.io/xugrid/examples/regridder_overview.html>`_) and planar coordinates (x, y).
```

imod/typing/grid.py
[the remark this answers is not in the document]
luitjansl: indeed, done.

JoerivanEngelen (commented):
Thanks for implementing this! Have you tested if it actually resulted in performance gains by regridding a simulation (e.g. one of the examples)? That would be worthwhile to do; feel free to make a separate issue if it hampers the merging of this PR too much.

My comments mostly concern documentation. Functionally, I think you have pretty much everything covered, the two things you should pay attention to are the hashing based on layer coordinates (as mentioned by @Huite), and the storing hashes of the target grid. For the latter it could be that I overlooked a usecase, and that it is useful to have.

Concerning documentation, we should make sure the RegriddingWeightsCache is included in the docs, as it is part of the public API. This should be added in the ``docs/mf6.rst`` to expose their docstrings. I see we do not include any of the utilities in there now, so we should start to pay more attention to that.

imod/typing/grid.py
luitjansl: done

imod/mf6/utilities/regrid.py
```diff
@@ -107,11 +91,31 @@ def get_regridder(
...
         regridder_class = self.__get_regridder_class(regridder_type)
 
-        if not self.__has_regridder(regridder_class, method):
-            self.__create_regridder(regridder_class, method)
+        if "layer" not in source_grid.coords and "layer" in target_grid.coords:
+            target_grid = target_grid.drop_vars("layer")
+            
+        source_hash  = get_grid_geometry_hash(source_grid)
+        target_hash  = get_grid_geometry_hash(target_grid) 
+        key = (source_hash, target_hash, regridder_class)
```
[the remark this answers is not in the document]
luitjansl: The regridder cache is accesible to users. When regridding a model or a simulation, the user doesn;t need to interact with it, but when regridding a package, the cache is a function argument. So users can create one in their own scripts and use it as they please. The target hash can be used to prevent misuse

imod/typing/grid.py
JoerivanEngelen: We could try and see if xxhash provides any significant performance gains for us. 

Installation wise:
[The python bindings are built to Windows, Linux, and Mac on conda-forge](https://anaconda.org/conda-forge/python-xxhash/). The [feedstock has two maintainers](https://github.com/conda-forge/python-xxhash-feedstock), but luckily they are very active.  So I don't foresee it making our installation process much harder than it already is.

examples/mf6/different_ways_to_regrid_models.py
luitjansl: done

JoerivanEngelen:
FYI: I made a issue about the public API https://github.com/Deltares/imod-python/issues/936

luitjansl:
> Thanks for implementing this! Have you tested if it actually resulted in performance gains by regridding a simulation (e.g. one of the examples)? That would be worthwhile to do; feel free to make a separate issue if it hampers the merging of this PR too much.

On the example "hondsrug_regridding", regridding takes  27.9 seconds with this approach, and 27.7 seconds with the approach taken on master.  I think it's not big enough to showcase the advantages of this method. We should try with a bigger model.

imod/typing/grid.py
luitjansl: let's first get it working like this, and measure if hashing is a time sink for large models. If it is, then we can try xxhash.

imod/typing/grid.py
```diff
@@ -316,3 +316,23 @@ def get_spatial_dimension_names(grid: xu.UgridDataArray) -> list[str]:
...
+    result = hash(pickle.dumps(grid["x"].values))
+    result += hash(pickle.dumps(grid["y"].values))
+    return result
+
+
+@typedispatch
+def get_grid_geometry_hash(grid: xu.UgridDataArray) -> int:
+    result = hash(pickle.dumps(grid.ugrid.grid.node_x))
+    result = +hash(pickle.dumps(grid.ugrid.grid.node_y))
+    result = +hash(pickle.dumps(grid.ugrid.grid.node_face_connectivity))
```
Huite, line 332: I'm not sure this the appropriate way of hashing these objects.
How likely is is that we get collisions?

For example, this is true:

```python
assert hash(100) + hash(1) == hash(101)
```
Would it be better to do this instead?

```python
hash_x = hash(pickle.dumps(grid.ugrid.grid.node_x)
hash_y = hash(pickle.dumps(grid.ugrid.grid.node_y)
hash_connectivity = hash(pickle.dumps(grid.ugrid.grid.node_face_connectivity)
return hash((hash_x, hash_y, hash_connectivity))
```
```

Comment (Huite) [reply imod/typing/grid.py:332]: Also the assignment operator is wrong!

```python
result += hash(...)
# versus
result = +hash(...)
```

The last is just assigning the value of to `result`.

luitjansl: done (both for structured and unstructured)

JoerivanEngelen (approved):

PR #941, opened 2024-03-28
Release 0.16.0

JoerivanEngelen:
# Description
Release 0.16.0 version, when this is merged we can run the new release pipeline.

### DIFF
--- a/docs/api/changelog.rst
+++ b/docs/api/changelog.rst
@@ -7,14 +7,13 @@ The format is based on `Keep a Changelog`_, and this project adheres to
 `Semantic Versioning`_.
 
 
-[Unreleased]
-------------
+[0.16.3] - 2024-03-29
+---------------------
 
 Fixed
 ~~~~~
-- Incorrect validation error ``data values found at nodata values of
-  idomain`` for boundary condition packages with a scalar coordinate
-  not set as dimension.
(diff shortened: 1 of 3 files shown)


docs/api/changelog.rst
```diff
@@ -39,22 +43,27 @@ Added
...
+  methods. `See the documentation
+  <https://deltares.github.io/imod-python/api/generated/logging/imod.logging.html>`_
   how to activate logging.
 - added :func:`imod.data.hondsrug_simulation` and
   :func:`imod.data.hondsrug_crosssection` data.
-- simulations and models that include a lake package now raise an exception on clipping, partitioning or regridding. 
+- simulations and models that include a lake package now raise an exception on
+  clipping, partitioning or regridding. 
 
 Changed
```
luitjansl, line 68: some of these changelog messages are only clear to imod-python developers, I think. Maybe we give too much detail?

luitjansl (approved):
looks good, 1 comment

docs/api/changelog.rst
JoerivanEngelen: They are indeed quite technical, but I think they are clear enough: We mention the specfic function/class/method in the add/fix/change. When users notice weird behavior in a specific function or class, they can CTRL+F to find a version number that changed something to this function.

PR #947, opened 2024-03-31
Issue #946: Validate >= instead of > for river bottom

### LINKED ISSUE #946: River validation: bottom_elevation should be >= bottom
I'm currently getting this validation error:

```
Simulation validation status:
	* GWF_1 model:
		* riv package:
			* bottom_elevation:
				* not all values comply with criterion: > bottom
```

In a model where I've set the bottom_elevation equal to the layer bottom.
In this specific case, I'm doing it simulate inflow along a vertical seepage/infiltration face.

If you disable this check, MODFLOW 6 will still run. See:

https://github.com/MODFLOW-USGS/modflow6/blob/23306af3a639b873e5eefd6b7dfcfa0fd8c6f5b6/src/Model/GroundWaterFlow/gwf-riv.f90#L253

```fortran
      ! -- accumulate errors
      if (rbot < bt .and. this%icelltype(node) /= 0) then
        write (errmsg, fmt=fmtriverr) i, rbot, bt
        call store_error(errmsg)
      end if
```

It errors if the river bottom `rbot` is below `bt`, but not if it's equal.
There is also a reasonable use case for this.

PR is incoming.

Huite:
Fixes #946

### DIFF
--- a/docs/api/changelog.rst
+++ b/docs/api/changelog.rst
@@ -6,6 +6,13 @@ All notable changes to this project will be documented in this file.
 The format is based on `Keep a Changelog`_, and this project adheres to
 `Semantic Versioning`_.
 
+[Unreleased]
+------------
+
+Fixed
+~~~~~
+- No ``ValidationError`` thrown anymore in :class:`imod.mf6.River` when
+  ``bottom_elevation`` equals ``bottom`` in the model discretization.
 
 [0.16.0] - 2024-03-29
 ---------------------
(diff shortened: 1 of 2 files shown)


JoerivanEngelen (approved):
Thanks for the fix!

I thought I had different experiences with Modflow 6 throwing me validation errors when ``bottom_elevation`` equals ``bottom``, but as you point out, that cannot be the case. I viewed the git blame and it has been that way for at least 4 years, so the validation was set too strict here.

I took the liberty of adding some text to the changelog as well, as this fix is relevant to users.

PR #957, opened 2024-04-04
Issue #536 misleading well message

### LINKED ISSUE #536: Misleading error in WEL package check during simulation.write()
In GitLab by @betsyromero on Sep 12, 2023, 19:48

In the checks done for the well package (imod.mf6.Well) in:

```python
simulation.write(modeldir, binary = False)
```

If the wells coordinates are outside the model area, the error obtained is:

`One or more well(s) are completely invalid due to minimum conductivity and thickness constraints.`

That error is misleading, as that is not the cause of the error.

Add write validation to check on wells outside of the model grid. This situation should produce an error: "one or more wells are outside of the model domain"

luitjansl:
Fixes #536 

# Description
This PR solves a bug. When a well was added outside of the domain, this resulted in an error message stating that the 
well was not added to the model because it violated constraints of minimum thickness / minimum k. The real reason, that the well was outside of the domain, was not mentioned. 
Now a separate check has been added that checks if the wells are inside the domain, and if not it give an error message
stating this, as well as the coordinates of the first well outside of the domain. 

#d Checklist

- [X] Links to correct issue
- [X] Unit tests were added

### DIFF
--- a/imod/prepare/wells.py
+++ b/imod/prepare/wells.py
@@ -47,7 +47,7 @@ def locate_wells(
     top: Union[xr.DataArray, xu.UgridDataArray],
     bottom: Union[xr.DataArray, xu.UgridDataArray],
     k: Union[xr.DataArray, xu.UgridDataArray, None],
-    throw_if_well_not_placed: bool = False
+    throw_if_well_not_placed: bool = False,
 ):
     if not isinstance(top, (xu.UgridDataArray, xr.DataArray)):
         raise TypeError(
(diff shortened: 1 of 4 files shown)


luitjansl:
some tests outside mf6 failed, back to the drawing board

imod/prepare/wells.py
```diff
@@ -47,6 +47,7 @@ def locate_wells(
     top: Union[xr.DataArray, xu.UgridDataArray],
     bottom: Union[xr.DataArray, xu.UgridDataArray],
     k: Union[xr.DataArray, xu.UgridDataArray, None],
+    throw_if_well_not_placed: bool = False,
```
JoerivanEngelen, line 50: Maybe just name this ``validate``? I see further down:

```
    if not np.array_equal(xy_bottom["index"], index):
        raise ValueError("bottom grid does not match top grid")
    if k is not None and not np.array_equal(xy_k["index"], index):
        raise ValueError("k grid does not match top grid")
```

We could set this to:

```
    if validate and not np.array_equal(xy_bottom["index"], index):
        raise ValueError("bottom grid does not match top grid")
    if validate and (k is not None) and not np.array_equal(xy_k["index"], index):
        raise ValueError("k grid does not match top grid")
```

JoerivanEngelen (commented):
Looks good, I have one comment:
Shouldn't the default setting for ``assign_wells`` and ``locate_wells`` not be True?
I think it is better to be restrictive by default, and allow the user to set less restrictive setting if they know what they are doing. This is how we implemented the ValidationSchemes as well. (And this is kind of an input validation)

I don't really mind the breaking change, as long as we advice users how to turn off this validation in the error message.

imod/prepare/wells.py
luitjansl: done

JoerivanEngelen (approved):

PR #973, opened 2024-04-18
Issue #950 Distribute conductances

### LINKED ISSUE #950: Create utility functions to distribute conductances
See this report from Sweco describing different methods to allocate cells with depth and distribute conductances over depth in iMOD5. Nice summary of requirements for this story!

[20240129 Conductanceverdeling iMOD5.docx](https://github.com/Deltares/imod-python/files/14222631/20240129.Conductanceverdeling.iMOD5.docx)

I translated the list of settings in iMOD5.6 briefly below. We could create a set of utility functions to support this, so that users can mix and match what they'd like.

Distribution conductance with depth:

* "default" iMOD5 option whatever that means?
* equally over layers
* weighted by layer thicknesses
* weighted by protruded layer thicknesses
* weighted by layer transmissivity
* weighted by horizontal hydraulic conductivity

JoerivanEngelen:
Fixes #950

# Description
Adds the following things:
* Adds utility functions to distribute conductances over depth, exposed to public API
* Moves allocation tests to test_topsystem.py file, so cases can be shared with the distribute conductances test.
* Moves cases to separate source file, do not use objects anymore to group cases, but function name prefixes. Furthermore tags are used to use tag which cases are used for RIV, DRN, GHB, and RCH tests. This saves quite some duplicate code.
* Move code ``basic_dis`` fixture to function ``make_basic_dis``, which is used to also make a 4 layered discretization fixture.
* Move duplicate code in allocation source file to ``_enforce_layered_top``

The only thing still missing is some extra options to distribute conductances for the Drain package. In this PR, it is not possible to use the crosscut thicknesses and correction factors yet. I'll make a separate task for that to not make the PR too overwhelming.

# Checklist

- [x] Links to correct issue
- [x] Unit tests were added


imod/prepare/topsystem/allocation.py
```diff
@@ -134,6 +134,13 @@ def _is_layered(grid: GridDataArray):
...
+def _enforce_layered_top(top: GridDataArray, bottom: GridDataArray):
+    if _is_layered(top):
+        return top
+    else:
+        return create_layered_top(bottom, top)
+
+
 @enforced_dim_order
 def _allocate_cells__stage_to_riv_bot(
     top: GridDataArray,
```
luitjansl, line 146: for this function it would help to know which arrays can be time-dependent and which ones are constant.

luitjansl (commented):
Looks good, and it contains nice tests. However, most test grids are simple in the sense that the layers are all horizontal and flat. It could be worth it to add some unit-tests for the _distribute_weights__by_XXX functions using grids with non-flat top and bottom arrays, so that cells will share a vertical cellface with cells in different layers, or which have inactive cells here and there.

imod/prepare/topsystem/allocation.py
JoerivanEngelen: Extended the docstrings with this information

JoerivanEngelen:
> Looks good, and it contains nice tests. However, most test grids are simple in the sense that the layers are all horizontal and flat. It could be worth it to add some unit-tests for the _distribute_weights__by_XXX functions using grids with non-flat top and bottom arrays, so that cells will share a vertical cellface with cells in different layers, or which have inactive cells here and there.

Inactive model cells should already be accounted for in ``allocated``.
I extended the test cases with cases for cells where the first up to the third layer are active, where the third layer only is active, and where only the first and the third layer are active.

luitjansl (approved):

PR #979, opened 2024-04-22
Issue #974 distribute drain conductance

### LINKED ISSUE #974: Extend distribution of conductances options for drains
#973 Adds all kinds of distribution options for different packages. In developing that PR I didn't realize yet that the options for the DRN can be extended, when we assume drains are draining all the way up to the surface elevation. In that case the drain elevation can be treated similar as the river bottom_elevation in the computation of distribution weights. So the following distributing options can be added:

- crosscut thickness
- crosscut transmissivity
- corrected transmissivity

JoerivanEngelen:
Fixes #974

# Description
This PR adds the options to distribute drain conductances.

* Adds ``by_crosscut_thickness``, ``by_crosscut_transmissivity``, ``by_corrected_transmissivity`` options to Drain package
* Make boundary conditions top and bottom arguments optional methods
* Change order of arguments conductance distribution functions

# Checklist

- [x] Links to correct issue
- [x] Unit tests were added

### DIFF
--- a/imod/tests/test_prepare/test_topsystem.py
+++ b/imod/tests/test_prepare/test_topsystem.py
@@ -136,7 +136,7 @@ def test_distribute_drn_conductance(
     conductance = zeros_like(elevation) + 1.0
 
     actual_da = distribute_drn_conductance(
-        option, allocated, conductance, top, bottom, k
+        option, allocated, conductance, top, bottom, k, elevation
     )
     actual = take_first_cell_in_xy_plane(actual_da)
 
--- a/imod/tests/test_prepare/test_topsystem.py
+++ b/imod/tests/test_prepare/test_topsystem.py
@@ -111,7 +111,7 @@ def test_distribute_riv_conductance(
     conductance = zeros_like(bottom_elevation) + 1.0
 
     actual_da = distribute_riv_conductance(
-        option, allocated, conductance, top, bottom, stage, bottom_elevation, k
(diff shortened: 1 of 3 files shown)


imod/tests/test_prepare/test_topsystem.py
```diff
@@ -136,7 +136,7 @@ def test_distribute_drn_conductance(
     conductance = zeros_like(elevation) + 1.0
 
     actual_da = distribute_drn_conductance(
-        option, allocated, conductance, top, bottom, k
+        option, allocated, conductance, top, bottom, k, elevation
     )
     actual = take_first_cell_in_xy_plane(actual_da)
```
luitjansl, line 141: you only compare "actual" and "expected" in the first cell.  So it would make the tests more strict if arrays like "bottom" and "elevation" have a different value in the first cell than in the rest of the array. Otherwise the tests would still pass even if something went wrong in the indexing.

luitjansl (approved):
1 comment ( but approved in advance)

imod/tests/test_prepare/test_topsystem.py
JoerivanEngelen: By doing this, we would mainly test if the testing utility function ``take_first_cell_in_xy_plane`` does what it's supposed to do. I think the it adds a quite some complexity to the test cases for relatively little gains. I think it is better to allocate our constrained resources to fixing other issues.

PR #981, opened 2024-04-23
Issue #964 from imod5 classmethbod DIS

### LINKED ISSUE #964: Add converter for iMOD5 Discretization package
If architecture as described in #961 is agreed on, we can create a converter for the DIS package.
For the Structured Discretization, the target grid is the smallest grid in the package. 
Therefore the method ``from_imod5_package`` needs to be overloaded:

```python
class StructuredDiscretization:
...
    @classmethod
    def from_imod5_package(cls, **data):
        # Find smallest grid in data to regrid to
        target = find_smallest_grid(data)
        # Regrid to one consistent dataset
        dataset = cls._regrid_like(target, **data)

        # Add logic to prepare the dataset in the following way here:
        # IBOUND -1  -> IDOMAIN = 1
        # IBOUND -1 cells will be converted to CHD cells
        dataset["idomain"] = dataset["idomain"].where(dataset["idomain"] > 0, 0)
        # Thickness <= 0 -> IDOMAIN = -1
        thickness = dataset["top"] - dataset["bottom"]
        dataset["idomain"] = dataset["idomain"].where(thickness > 0, -1)
        # TOP 3D -> TOP 2D
        dataset["top"] = dataset["top"].sel(layer=1, drop=True)

        return cls(**dataset)
```

JoerivanEngelen:
Fixes #964 

# Description
Implements from_imod5 classmethod to StructuredDiscretization classmethod.

NOTE: I had to modify the default regridding method of "idomain" as the "mode" method appears to have a bug with perfectly aligned structured grids.
This bug requires further investigation. Will post an issue if found.

# Checklist

- [x] Links to correct issue
- [x] Unit tests were added


imod/mf6/dis.py
```diff
@@ -151,3 +160,81 @@ def _validate(self, schemata, **kwargs):
...
+        -------
+        Modflow 6 StructuredDiscretization package.
+
+        """
+        data = {
+            "idomain": imod5_data["bnd"]["ibound"].astype(np.int32),
+            "top": imod5_data["top"]["top"],
+            "bottom": imod5_data["bot"]["bottom"],
+        }
+        # TODO: Add validation
```
luitjansl, line 202: you can create an issue for validation, and remove this comment block.

imod/mf6/dis.py
```diff
@@ -151,3 +160,81 @@ def _validate(self, schemata, **kwargs):
...
+        # It is possible that top and bottom grids are on different grids, then
+        # this doesn't work:
+        # assert np.all(imod5_data["top"]["top"][1:].data == imod5_data["bot"]["bottom"][:-1].data)
+        # assert np.isin(np.unique(imod5_data["bnd"]["ibound"]), [-1, 0, 1])
+
+        target_grid = get_smallest_target_grid(*data.values())
+
+        regridder_settings = cls._regrid_method
```
luitjansl, line 210: make a deepcopy of cls._regrid_,method to avoid changing the defaults in this assignment. This is something that must be done at more locations. probably every implementation of "get_regrid_methods"should do this.

imod/mf6/dis.py
```diff
@@ -151,3 +160,81 @@ def _validate(self, schemata, **kwargs):
...
+        if regridder_types is not None:
+            regridder_settings.update(regridder_types)
+
+        # TODO: are grid arguments really used in RegridderWeightsCache after
+        #   initialization? Don't really seem to be...
+        regrid_context = RegridderWeightsCache(data["idomain"], target_grid)
```
luitjansl, line 216: are the top and bot arrays all defined on  the same grid as idomain? The RegridderWeightsCache assumes we have 1 source grid and 1 target grid, and it stores weights for different regridding methodsfor this source and target.

imod/mf6/package.py
```diff
@@ -78,24 +79,7 @@ def sel(self):
         )
 
     def _valid(self, value):
```
luitjansl, line 81: "self" is not used

imod/mf6/utilities/imod5_converter.py
```diff
@@ -0,0 +1,25 @@
+import numpy as np
+
+from imod.typing.grid import GridDataArray, full_like
+
+
+def convert_ibound_to_idomain(ibound: GridDataArray, thickness: GridDataArray):
```
luitjansl, line 6: GridDataArray means it can be structured or unstructured. If we test only for structured, it may be better to specify xr.DataArray here (to not suggest it is more general than it really is)

imod/mf6/dis.py
```diff
@@ -88,7 +97,7 @@ class StructuredDiscretization(Package, IRegridPackage):
     _regrid_method = {
         "top": (RegridderType.OVERLAP, "mean"),
         "bottom": (RegridderType.OVERLAP, "mean"),
-        "idomain": (RegridderType.OVERLAP, "mode"),
+        "idomain": (RegridderType.OVERLAP, "sum"),
```
luitjansl, line 100: here, you could add a TODO to remind us to put "mode" back for idomain regridding

luitjansl (commented):
looks good, some comments

imod/mf6/dis.py
JoerivanEngelen: Very good catch! Implementing this elsewhere is worthy of its own issue. Created an issue for this https://github.com/Deltares/imod-python/issues/1000

PR #992, opened 2024-04-25
Issue #991 Only evaluate flip_svat_y for flipping row indices

### LINKED ISSUE #991: Reversed y-coord indices in grid_mapping method of MeteoMapping class of MetaSWAP model
In method, the flip_meteo_y variable is evaluated based on the y-coord of the meteogrid. But the actual used bounds are always monotonic increasing. This results in a faulty evaluation on flipped y-axis. 

Fix: Assume meteo_y not to be flipped, only evaluate flip_svat_y.

HendrikKok:
Fixes #991

# Description

# Checklist

- [x] Links to correct issue
- [x] Unit tests were added

### DIFF
--- a/docs/api/changelog.rst
+++ b/docs/api/changelog.rst
@@ -32,7 +32,7 @@ Fixed
   error message stating a well is outside of the domain.
 - When importing data from a .prj file, the multipliers and additions specified for
   ipf and idf files are now applied
-- Compute correct indices for svat - meteogrid mapping, in imod.msw.MeteoMapping
+- Fix bug where y-coords were flipped in :class:`imod.msw.MeteoMapping`
   
 [0.16.0] - 2024-03-29
 ---------------------
--- a/docs/api/changelog.rst
+++ b/docs/api/changelog.rst
@@ -32,6 +32,7 @@ Fixed
   error message stating a well is outside of the domain.
 - When importing data from a .prj file, the multipliers and additions specified for
   ipf and idf files are now applied
+- Compute correct indices for svat - meteogrid mapping, in imod.msw.MeteoMapping
   
 [0.16.0] - 2024-03-29
 ---------------------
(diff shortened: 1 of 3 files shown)


JoerivanEngelen (commented):
Inferring from the test, this seems to do the right thing. However, it appears to the exact opposite thing of the fix proposed in the issue. Could you elaborate?

This fix is also important to users. Could you add a text to the changelog?
It's under ``docs/api/changelog.rst``.

HendrikKok:
> Inferring from the test, this seems to do the right thing. However, it appears to the exact opposite thing of the fix proposed in the issue, where you mention only flip_svat_y should be evaluated, whereas the PR removes exactly that. Could you elaborate?
> 
> This fix is also important to users. Could you add a text to the changelog? It's under `docs/api/changelog.rst`.

JoerivanEngelen:
> Yes, good point and that’s indeed my bad. The initial proposed fix was based on the typical Dutch case with positive increasing x- values and negative for y, for meteo- and svat array. Since this fix made the current test fail, I looked more closely to the logic.
> 
> In my opinion, the only reason to flip the indices is if the edge array (always monotonic increasing) is flipped relative to the array-coords (potentially monotonic decreasing). Therefore, there is no reason the evaluate the array-coords of the svat-array at all.

Yeah the part dealing with y-coords is a bit sketchy. I think the main culprit is the MetaSWAP IO manual stating somewhere that y-coords should be monotonically increasing for a specific package (I think it was the IDF mapping), and I doubt whether that is really the case. I think the main use case is decreasing y-coords. So we could throw a ValidationError if the input has increasing y coords, as we do with the Modflow6 module. Indeed the SVAT logic can be removed.

docs/api/changelog.rst
```diff
@@ -32,6 +32,7 @@ Fixed
   error message stating a well is outside of the domain.
 - When importing data from a .prj file, the multipliers and additions specified for
   ipf and idf files are now applied
+- Compute correct indices for svat - meteogrid mapping, in imod.msw.MeteoMapping
```
JoerivanEngelen, line 35: To render this nicely:

```suggestion
- Fix bug where y-coords were flipped in :class:`imod.msw.MeteoMapping`
```

JoerivanEngelen (approved):
Approving in advance, could you specify the sentence a bit further? I have a suggestion to improve this.

Issue #990, opened 2024-04-25
Extend MyPy checks to include more modules.


JoerivanEngelen:
MyPy is currently only applied for the ``mf6`` and ``logging`` modules. 

We should extend this to other modules:
- [ ] ``msw``
- [ ] ``prepare``
- [ ] ``util``
- [ ] ``typing``
- [ ] ``select``

PR #995, opened 2024-04-29
Issue #990 mypy

luitjansl:
Fixes #

# Description
Includes more source directories for mypy checking. Solves some of the issues that were detected by mypy.

# Checklist

- [X] Links to correct issue
- [ ] Unit tests were added

### DIFF
--- a/imod/msw/pkgbase.py
+++ b/imod/msw/pkgbase.py
@@ -18,6 +18,7 @@ class MetaSwapPackage(abc.ABC):
     """
 
     __slots__ = "_pkg_id"
+    _file_name = "filename not set"
 
     def __init__(self):
         self.dataset = xr.Dataset()
--- a/imod/prepare/spatial.py
+++ b/imod/prepare/spatial.py
@@ -948,7 +948,7 @@ def zonal_aggregate_raster(
     raster: xr.DataArray,
     resolution: float,
     method: Union[str, Callable],
-    chunksize: int = 1e4,
+    chunksize: int =  int(1e4),
 ) -> pd.DataFrame:
     """
     Compute a zonal aggregate of raster data for polygon geometries, e.g. a mean,
@@ -1031,7 +1031,7 @@ def zonal_aggregate_polygons(
     like: xr.DataArray,
     resolution: float,
     method: Union[str, Callable],
-    chunksize: int = 1e4,
+    chunksize: int = int(1e4),
 ) -> pd.DataFrame:
     """
     Compute a zonal aggregate of polygon data for (other) polygon geometries,
--- a/imod/prepare/spatial.py
+++ b/imod/prepare/spatial.py
@@ -948,7 +948,7 @@ def zonal_aggregate_raster(
     raster: xr.DataArray,
     resolution: float,
     method: Union[str, Callable],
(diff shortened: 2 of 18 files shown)


imod/util/structured.py
```diff
@@ -98,7 +98,7 @@ def _replace(
         out[replaceable] = value[indices[replaceable]]
         return out.reshape(a.shape)
 
-    if is_scalar(to_replace):
+    if isinstance(to_replace, int | float):
```
JoerivanEngelen, line 101: ``is_scalar`` does more than only testing if ``int`` or ``float``. For example a np.array(2.0) is also possible or xr.DataArray(3). What error does it print? Otherwise just ignore it with a docstring as done on other places.

imod/util/time.py
```diff
@@ -189,7 +189,7 @@ def _compose_timestring(
     types of available time objects.
     """
     if time == "steady-state":
-        return time
+        return str(time)
```
JoerivanEngelen, line 192: Please add comment that this is to avoid a mypy error. 

For the common observer, it makes sense that ``"steady-state"`` is always a string, so this string conversion appears trivial.

JoerivanEngelen (commented):
Looks good mostly, but some comments. The only thing I really want to see reverted is the change of ``is_scalar`` to a call to ``isinstance``. As that changes behavior.

imod/util/structured.py
luitjansl: the issue here is in fact line 104. On that line we use` np.isnan(to_replace)`, and `np.isnan` does not necessarily work on all the types that `to_replace ` could have. So in my initial solution I fixed this by limiting the options to int or float, for which `np.isnan` defenitely works. 
Now I reverted to leaving line 101 untouched (so using` is_scalar`)  and I added the `type:ignore` to line 104

imod/util/time.py
luitjansl: done

imod/msw/pkgbase.py
```diff
@@ -18,6 +18,7 @@ class MetaSwapPackage(abc.ABC):
     """
 
     __slots__ = "_pkg_id"
+    _file_name = "filename not set"
```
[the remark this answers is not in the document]
luitjansl: done

imod/prepare/spatial.py
```diff
@@ -948,7 +948,7 @@ def zonal_aggregate_raster(
     raster: xr.DataArray,
     resolution: float,
     method: Union[str, Callable],
-    chunksize: int = 1e4,
+    chunksize: int = int(1e4),
```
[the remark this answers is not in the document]
luitjansl: done

imod/prepare/spatial.py
```diff
@@ -1031,7 +1031,7 @@ def zonal_aggregate_polygons(
     like: xr.DataArray,
     resolution: float,
     method: Union[str, Callable],
-    chunksize: int = 1e4,
+    chunksize: int = int(1e4),
```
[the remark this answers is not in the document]
luitjansl: done

JoerivanEngelen (approved):

PR #1005, opened 2024-05-02
Add support for the fixed_cell option to the Recharge package

HendrikKok:
Add support for the ``fixed_cell`` option to the Recharge package. This option is relevant for phreatic models, not using 
the Newton formulation and model cells can become inactive. The preferred method 
for phreatic models is to use the Newton formulation, where cells remain active, 
and this option irrelevant.

# Checklist

- [ ] Links to correct issue
- [x] Unit tests were added


JoerivanEngelen (commented):
As discussed via a call:
- Could you update the changelog?
- Explain how this complexity differs from the SolutionPreset complexity levels. Does it do somewhat the same or something entirely different?

JoerivanEngelen (approved):

PR #1012, opened 2024-05-06
Add support for discharge scaling in drainage package

### LINKED ISSUE #1011: Add support for discharge scaling depth in drainage package
MODFLOW 6 supports drainage discharge scaling: when heads near the drainage elevation, they may be reduced by a scaling factor. See also the supplementary info chapter 3. 

Cubic smoothing is used to ensure continuous discharge derivatives when the Newton-Rhapson formulation is used, which helps convergence.

Huite:
Fixes #1011

Still a draft, this adds the drainage discharge scaling depth.

This is a rather pragmatic approach. We're currently only using the auxiliary variables for concentration, which are treated separately. The `_auxiliary_data = {"concentration": "species"}` cannot be meaningfully used, since this specific `scaling_depth` does not contain any additional dimensions.

Adding it to `period_data` ensures that it is written to the data files, but since it's optional, the presences must then be checked in `get_period_names`.

### DIFF
--- a/imod/mf6/boundary_condition.py
+++ b/imod/mf6/boundary_condition.py
@@ -303,7 +303,7 @@ class BoundaryCondition(Package, abc.ABC):
         if hasattr(self, "_auxiliary_data"):
             result.extend(get_variable_names(self))
 
-        return result
+        return [var for var in result if var in self.dataset.data_vars]
 
 
 class AdvancedBoundaryCondition(BoundaryCondition, abc.ABC):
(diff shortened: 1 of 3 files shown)


Huite:
Something to consider when dealing with the auxiliary arrays: MODFLOW 6 has no info in the .dat/.bin files to see which column is which. Instead, (I'm pretty sure) the meaning of the columns depends on the order of the auxiliary entry in the options block.

This means it's to coordinate the order of the variables when converting to a dictionary of arrays with the auxiliary entry written in the options block. Otherwise you end up with MF6 interpreting the concentration as the factor or vice versa.

PR #1014, opened 2024-05-06
Issue #1013 Make linting stricter

### LINKED ISSUE #1013: Make ruff checks more strict

Manangka:
Fixes #1013

# Description
This commit makes the ruff formatter and linter stricter.
When switching to ruff we had to ignore a lot of the linting rules to keep the build green.
Most of the ignored rules have been resolved in this commit

The rules have been grouped per commit. To help you in reviewing this you can go through the commits one by one to see related changes

# Checklist

- [x] Links to correct issue
- [ ] Unit tests were added


JoerivanEngelen (changes requested):
Thanks for the fixes! Most of them seem fine to me, I have one comment though.

For the examples online, it is nicer to do imports in their respective code block (i.e. a block of code starting from a ``# %%``) for the jupyter notebooks. For example, call ``import matplotlib.pyplot as plt`` in a code block where a plot is made, instead of on top of the script. In this way, users are less overwhelmed with info when they start going through the example. I see for some examples you have preserved this order with a ``# noqa: E402 comment``, and for others all imports are moved to the top. I think the first is the nicer option, but it still has a weird comment, entirely irrelevant to users; so only confusing them.

Ideally, I would like that we can make an exception to E402 for the whole example folder, as we know beforehand that we would like imports to not necessarily be on top. [I read that this is possible here](https://stackoverflow.com/questions/77680073/ignore-specific-rules-in-specific-directory-with-ruff). Could you try if you can make that work and revert the changes made to imports in the examples?

Manangka:
> Thanks for the fixes! Most of them seem fine to me, I have one comment though.
> 
> For the examples online, it is nicer to do imports in their respective code block (i.e. a block of code starting from a `# %%`) for the jupyter notebooks. For example, call `import matplotlib.pyplot as plt` in a code block where a plot is made, instead of on top of the script. In this way, users are less overwhelmed with info when they start going through the example. I see for some examples you have preserved this order with a `# noqa: E402 comment`, and for others all imports are moved to the top. I think the first is the nicer option, but it still has a weird comment, entirely irrelevant to users; so only confusing them.
> 
> Ideally, I would like that we can make an exception to E402 for the whole example folder, as we know beforehand that we would like imports to not necessarily be on top. [I read that this is possible here](https://stackoverflow.com/questions/77680073/ignore-specific-rules-in-specific-directory-with-ruff). Could you try if you can make that work and revert the changes made to imports in the examples?

Great suggestion. I reverted the changes to the examples related to E402. Furthermore i added an ignore of that rule to the example folder as suggested by the link you posted

JoerivanEngelen (approved):

PR #1018, opened 2024-05-07
Update ims options

Huite:
MODFLOW 6 has changed a number of options in the IMS package (since quite some time).

* This adds support for outer_csvfile and inner_csvfile instead of csv_output.
* This simplifies the no_ptc option: instead of a bool and a possible additional entry, just the entry is required. (In the IMS file, NO_PTC is equal to NO_PCT ALL, i.e. the ALL is an assumed default value.)
* This also adds support for the adaptive time stepping entry ats_outer_maximum_fraction.

All the deprecated solution keyword arguments have been removed from the examples.


imod/mf6/ims.py
```diff
@@ -419,7 +425,9 @@ def __init__(
             "scaling_method": scaling_method,
             "reordering_method": reordering_method,
             "print_option": print_option,
-            "csv_output": csv_output,
+            "outer_csvfile": outer_csvfile,
+            "inner_csvfile": inner_csvfile,
+            "ats_outer_maximum_fraction": ats_outer_maximum_fraction,
             "no_ptc": no_ptc,
         }
         # Make sure the modelnames are set as a variable rather than dimension:
```
JoerivanEngelen, line 433: I think we should have some validation to check if no_ptc equals"first" or "all". I don't wanna burden you further with defining a ValidationScheme for this, but could you add some code in the init that does:

```suggestion
            if setting.lower() in ["first", "all"]:
                return setting.lower()
            else:
                raise ValueError(
                    f"Received wrong setting for no_ptc. String should be one of ['first', 'all'], instead got {setting}"
                )
        # Make sure the modelnames are set as a variable rather than dimension:
```

(Modified from ``OutputControl._get_ocsetting``)

JoerivanEngelen (commented):
I have one comment to have some simple input validation, furthermore could you update the changelog? As the changes affect the public API.

imod/mf6/ims.py
Huite: I've added an OptionSchema instead. Should be generally useful.

JoerivanEngelen (approved):

PR #1021, opened 2024-05-07
Issue #1017 fix at first active

### LINKED ISSUE #1017: In topsystem allocation, at_first_active method allocates to all active model cells.
It should allocate only to cells in active boundary condition cells, not to every potential model active. See example visuals, where the left hand side allocates correctly (where river cells are defined), but on the right hand side ``at_first_active`` allocates everywhere.

I think I first didn't implement this as presumably this fixes itsself when broadcasting with ``rate.where(rch_allocated)`` or ``rate * rch_allocated``. Yet you never know what users do, fixing it is easy, and it makes for a confusing example.

image

This will alter the public API a bit: As ``allocate_rch_cells`` will now also require a ``rate`` argument. So this issue should be preferably solved before the next (0.17.0) release, when we release all topsystem allocation and distrubion functionality.

JoerivanEngelen:
Fixes #1017

# Description
This ensures no cells are allocated to cells without a topsystem assigned to them. I added one inactive cell to the test cases and test if it returns False

# Checklist

- [x] Links to correct issue
- [x] Unit tests were added

### DIFF
--- a/imod/tests/test_prepare/test_topsystem.py
+++ b/imod/tests/test_prepare/test_topsystem.py
@@ -21,7 +21,7 @@ def take_nth_cell_in_xy_plane(grid: GridDataArray, n: int) -> GridDataArray:
         return grid.values[:, n]
     else:
         return grid.values[:, n, n]
-    
+
 
 @parametrize_with_cases(
     argnames="active,top,bottom,stage,bottom_elevation",
--- a/imod/tests/test_prepare/test_topsystem.py
+++ b/imod/tests/test_prepare/test_topsystem.py
@@ -16,7 +16,21 @@ from imod.typing.grid import is_unstructured, zeros_like
 from imod.util.dims import enforce_dim_order
 
 
-def take_nth_cell_in_xy_plane(grid: GridDataArray, n: int) -> GridDataArray:
+def take_nth_layer_column(grid: GridDataArray, n: int) -> GridDataArray:
+    """
+    Parameters
+    ----------
+    grid: DataArray | UgridDataArray
+        grid to take values from. Must have dimensions (layer,y,x) for
+        structured and (layer,{face_dim}) for unstructured grids.
+    n: int
+        index number in the xy plane where layer column is taken.
+
+    Returns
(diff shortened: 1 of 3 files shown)


imod/tests/test_prepare/test_topsystem.py
```diff
@@ -37,15 +37,21 @@ def test_riv_allocation(
         option, active, top, bottom, stage, bottom_elevation
     )
 
-    actual_riv = take_first_cell_in_xy_plane(actual_riv_da)
+    actual_riv = take_nth_cell_in_xy_plane(actual_riv_da, 0)
+    empty_riv = take_nth_cell_in_xy_plane(actual_riv_da, 1)
 
     if actual_drn_da is None:
         actual_drn = actual_drn_da
```
luitjansl, line 44: actual_drn = None
is more explicit

luitjansl (approved):
looks good, some very minor comments.

imod/tests/test_prepare/test_topsystem.py
JoerivanEngelen: According to PEP8, ``is None`` is the preferred way to do things, as None is a singleton. See https://jaredgrubb.blogspot.com/2009/04/python-is-none-vs-none.html

imod/tests/test_prepare/test_topsystem.py
```diff
@@ -16,11 +16,11 @@
 from imod.util.dims import enforce_dim_order
 
 
-def take_first_cell_in_xy_plane(grid: GridDataArray) -> GridDataArray:
+def take_nth_cell_in_xy_plane(grid: GridDataArray, n: int) -> GridDataArray:
```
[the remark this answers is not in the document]
JoerivanEngelen: Done

imod/tests/test_prepare/test_topsystem.py
JoerivanEngelen: Oh sorry I misread your comment, that indeed makes the code more readable. I'll change that.

PR #1031, opened 2024-05-13
Issue #969 imod5 import recharge

### LINKED ISSUE #969: Add converter for iMOD5 RCH package
Needs to:
* Allocate planar recharge rates or constants to uppermost active cells.

Blocked by:
https://github.com/Deltares/imod-python/issues/949

We think boundary conditions always need to be assigned to a active cell, but this can be verified.

luitjansl:
Fixes #969

# Description
implements importing recharge packages from a project file. The recharge can be imported using a planar grid, 
in which case it assigns the recharge to the uppermost active cell for each column.
The recharge can also be imported from a non-planar (so fully 3d) grid in which case it is just regridded to the target grid. 

# Checklist
- [X] Links to correct issue
- [X] Unit tests were added


imod/typing/grid.py
```diff
@@ -398,3 +398,27 @@ def decorator(*args, **kwargs):
...
+) -> bool:
+    # Returns True if the grid is planar. It has then a layer coordinate with
+    # length 1 and value 0, or an empty layer coordinate axis, or no layer coordinate at all
+    if "layer" not in grid.coords:
+        return True
+    if grid["layer"].shape == ():
+        return True
+    if grid["layer"][0] == 0 and len(grid["layer"]) == 1:
+        return True
+    return False
```
JoerivanEngelen, line 414: Nice, I think a requirement for this one also that it is spatial_2D. Settings stored in xr.DataArrays/xr.Datasets without a dimension now also return True. 
For example:
```python
ds = xr.Dataset()
ds["save_flows"] = True

print(is_planar_grid(ds))
# Prints True, but should print False

```

So I suggest something like:

```suggestion
def is_planar_grid(
    grid: xr.DataArray | xr.Dataset | xu.UgridDataArray | xu.UgridDataset,
) -> bool:
    if not is_spatial_2D(grid):
         return False
    # Returns True if the grid is planar. It has then a layer coordinate with
    # length 1 and value 0, or an empty layer coordinate axis, or no layer coordinate at all
    if "layer" not in grid.coords:
        return True
    if grid["layer"].shape == ():
        return True
    if grid["layer"][0] == 0 and len(grid["layer"]) == 1:
        return True
    return False
```

imod/mf6/rch.py
```diff
@@ -152,3 +161,80 @@ def _validate(self, schemata, **kwargs):
...
+                )
+                for time in data["rate"]["time"].values:
+                    rch_rate.loc[time, :, :, :] = (
+                        data["rate"].sel(time=time).drop_vars("layer")
+                    )
+            else:
+                rch_rate.loc[:, :, :] = data["rate"].drop_vars("layer")
+
+            # remove rch from cells where it is not allocated
+            rch_rate = xr.where(is_rch_cell, rch_rate, np.nan)
```
JoerivanEngelen, line 237: I think we can better utilize xarray's broadcasting here:
- We do not have to load each timestep into memory and assign manually.
- Code can be reduced
- You are assigning the old ``data["rate"]`` instead of the regridded ``new_package_data["rate"]`` to ``rch_rate``

I might be naive, but I think we can reduce these data transformations to:

```suggestion

            # remove rch from cells where it is not allocated and broadcast over layers.
            rch_rate = new_package_data["rate"].where(is_rch_cell)
            rch_rate = enforce_dim_order(rch_rate)
```

imod/tests/test_mf6/test_mf6_rch.py
```diff
@@ -324,3 +326,118 @@ def test_clip_box(rch_dict):
...
+    data = deepcopy(imod5_dataset)
+
+    # make the first layer of the target grid inactive
+    target_grid = data["khv"]["kh"]
+    target_grid.loc[{"layer": 1}] = 0
+
+    # the input for recharge is on the second layer of the targetgrid
+    input_recharge = nan_like(target_grid)
+    input_recharge.loc[{"layer": 2}] = data["rch"]["rate"].sel(
+        layer=1, method="nearest"
```
JoerivanEngelen, line 424: Please add some comment why you had to choose "nearest" here (or that you don't know but it was required for some reason)

imod/tests/test_typing/test_typing_grid.py
```diff
@@ -64,3 +69,56 @@ def test_enforce_dim_order__unstructured(basic_unstructured_dis):
 
     assert actual.dims == ibound.dims
     assert isinstance(actual, type(ibound))
+
+
+def test_is_transient(basic_dis, basic_unstructured_dis):
```
JoerivanEngelen, line 74: I don't think this name is appropriate: You don't seem to call ``is_transient`` here, I only see a call to ``enforce_dim_order`` 

Some leftover copypasta?

JoerivanEngelen (changes requested):
I have some comments!

imod/typing/grid.py
luitjansl: done, nice catch.

imod/tests/test_typing/test_typing_grid.py
luitjansl: indeed, I removed it.

imod/mf6/rch.py
luitjansl: nice, much better

imod/mf6/rch.py
```diff
@@ -152,3 +161,68 @@ def _validate(self, schemata, **kwargs):
...
+        new_package_data = _regrid_package_data(
+            data, target_grid, regridder_settings, regrid_context, {}
+        )
+
+        # if rate has only layer 0, then it is planar.
+        if is_planar_grid(data["rate"]):
+            # create an array indicating in which cells rch is active
+            is_rch_cell = allocate_rch_cells(
+                ALLOCATION_OPTION.at_first_active,
+                target_grid,
```
JoerivanEngelen, line 219: The ``active`` grid (``idomain == 1``) should be inserted here. This comes from the regridded discretization package. So you therefore require an extra argument to the ``from_imod5_data`` method, namely the regridded discretization package.

For an example of the API of the ``allocate_<pkg>_cells``, see: https://deltares.github.io/imod-python/user-guide/09-topsystem.html#allocate-river-cells

imod/tests/test_mf6/test_mf6_rch.py
```diff
@@ -324,3 +326,118 @@ def test_clip_box(rch_dict):
     selection = rch.clip_box(x_min=10.0, x_max=20.0, y_min=10.0, y_max=20.0)
     assert selection["rate"].dims == ("y", "x")
     assert selection["rate"].shape == (1, 1)
+
+
+@pytest.mark.usefixtures("imod5_dataset")
+def test_planar_rch_from_imod5_constant(imod5_dataset, tmp_path):
+    data = deepcopy(imod5_dataset)
+    target_grid = data["khv"]["kh"]
```
JoerivanEngelen, line 334: We do not test if the regridding works now... I don't think we need to that for each package, but having one where it is tested helps I think. Could you add a test where you regrid this case to a very coarse grid?

imod/tests/test_mf6/test_mf6_rch.py
luitjansl: it turned out (as you suggested) that the deepcopy statements at the beginning of these tests ( top copy the fixture) does not work as expected: the fixture data is still modified by the tests. This is why this test needed the method=nearest to succeed when the test was performed along with the other tests. One of the earlier tests tampered with the layer coordinate. I have solved this now by putting the layer coordinate back to its original value in this test. But we should make a true deepcopy of the test data at some point.

imod/mf6/rch.py
luitjansl: done

imod/tests/test_mf6/test_mf6_rch.py
luitjansl: in the latest version of this changeset we use the discretization package as the target grid, and it is much finer than the  target_grid = data["khv"]["kh"] that was used earlier. So now it is regridding.
That aside, xugrid does as far as I know the same operation, whether the target grid matches the input grid exactly or not. So it goes through the same code path no matter what target grid we provide.

imod/tests/test_typing/test_typing_grid.py
```diff
@@ -64,3 +69,43 @@ def test_enforce_dim_order__unstructured(basic_unstructured_dis):
...
+        # set layer coordinates as  present and not  empty or 0
+        bottom_layer = bottom_layer.expand_dims({"layer": [9]})
+        assert not is_planar_grid(bottom_layer)
+
+        # set layer coordinates as  present and   0
+        bottom_layer.coords["layer"].values[0] = 0
+        assert is_planar_grid(bottom_layer)
+
+
+def test_is_transient_grid(basic_dis, basic_unstructured_dis):
```
JoerivanEngelen, line 95:
```suggestion
def test_is_transient_data_grid(basic_dis, basic_unstructured_dis):
```

imod/mf6/rch.py
```diff
@@ -152,3 +164,69 @@ def _validate(self, schemata, **kwargs):
 
     def get_regrid_methods(self) -> Optional[dict[str, Tuple[RegridderType, str]]]:
         return self._regrid_method
+
+    @classmethod
+    def from_imod5_data(
+        cls,
+        imod5_data: dict[str, dict[str, GridDataArray]],
+        discretization_package: StructuredDiscretization,
```
JoerivanEngelen, line 172: I think this can be shortened to:

```suggestion
        dis_pkg: StructuredDiscretization,
```

As ``DIS`` is an abbreviation used by MODFLOW6 and ``pkg`` is a very common abbreviation for "package".

imod/mf6/rch.py
```diff
@@ -152,3 +164,69 @@ def _validate(self, schemata, **kwargs):
...
+        ----------
+        imod5_data: dict
+            Dictionary with iMOD5 data. This can be constructed from the
+            :func:`imod.formats.prj.open_projectfile_data` method.
+        target_grid: GridDataArray
+            The grid that should be used for the new package. Does not
+            need to be identical to one of the input grids.
+        regridder_types: dict, optional
+            Optional dictionary with regridder types for a specific variable.
+            Use this to override default regridding methods.
```
JoerivanEngelen, line 193: Please update docstring with extra ``dis_pkg``/``discretization_package`` argument.

imod/tests/fixtures/backward_compatibility_fixture.py
```diff
@@ -8,8 +8,13 @@
...
     data = imod.data.imod5_projectfile_data(tmp_path)
-    _load_imod5_data_in_memory(data[0])
-    return data[0]
+    data = data[0]
+    # Fix data for ibound  as it contains floating values like 0.34, 0.25 etc.
+    ibound = data["bnd"]["ibound"]
+    ibound = ibound.where(ibound <= 0, 1)
+    data["bnd"]["ibound"] = ibound
+    _load_imod5_data_in_memory(data)
+    return data
```
JoerivanEngelen, line 17: This cleanup also occurs in the tests ``test_mf6_dis.py``, please remove the ibound cleanup from these tests.

Also: You might get some increased performance by first loading the data into memory and then cleaning up ibound.

JoerivanEngelen (approved):
Some minor comments, approving in advance

imod/mf6/rch.py
luitjansl: done

imod/mf6/rch.py
luitjansl: done

PR #1035, opened 2024-05-14
Issue #1034 Add Stricter MyPy settings

### LINKED ISSUE #1034: Make MyPy use strict settings

Manangka:
Fixes #1034

# Description
This commit adds the MyPy stricter setting to the PyProject.toml file and also fixes some more MyPy errors.

The result of this settings is that instead of rules being disabled by default we know have to disable them explicitly.
This make it visible which rules we are not applying and we still need to fix. 

The ideal situation would be to have no mypy rules at all in tjhe project files which means we are using complying to all MyPy rules.
The road to MyPy strict settings is described by  https://mypy.readthedocs.io/en/stable/existing_code.html#introduce-stricter-options

# Checklist

- [x] Links to correct issue
- [ ] Unit tests were added

### DIFF
--- a/imod/mf6/hfb.py
+++ b/imod/mf6/hfb.py
@@ -408,11 +408,16 @@ class HorizontalFlowBarrierBase(BoundaryCondition, ILineDataPackage):
                 )
             )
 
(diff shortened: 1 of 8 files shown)


JoerivanEngelen (approved):
Looks good, thanks!

PR #1048, opened 2024-05-21
Issue #1025 Shorten duration tests on teamcity with xdist

### LINKED ISSUE #1025: Unittests take too long
The unittests take too long; locally on my machine ~15 minutes but on TeamCity they approach 45 minutes. 

[Looking at this example](https://dpcbuild.deltares.nl/buildConfiguration/iMOD6_IMODPython_Windows_UnitTests/3766528?buildTab=tests&orderBy=duration&order=desc&expandTestHistoryChartSection=true), I observe the following things:

- `imod.tests.test_mf6.test_mf6_regrid_model.test_model_regridding_can_validate` and `imod.tests.test_mf6.test_mf6_regrid_model.test_model_regridding_can_skip_validation` both take nearly 1 minute. These are the longest tests so far in the pipeline
- There are a lot of tests for splitting and partitioning. We should investigate if we can coarsen grid for these test cases to speed up times.

If this is not doable, we can consider creating a separate pipeline for all the multimodel tests, which can run in parallel.

Alternatively: We can adapt our test strategy and include certain slower tests in a nightly pipeline. In this way changes which break things are not directly identified, but still be traced back to the day of a breaking commit.

Manangka:
Fixes #1025

# Description
Test to see if xdist will improve our pipeline build time

# Checklist

- [x] Links to correct issue
- [ ] Unit tests were added

### DIFF
--- a/pixi.lock
+++ b/pixi.lock
@@ -80,6 +80,7 @@ environments:
       - conda: https://conda.anaconda.org/conda-forge/noarch/editables-0.5-pyhd8ed1ab_0.conda
       - conda: https://conda.anaconda.org/conda-forge/linux-64/eigen-3.4.0-h00ab1b0_0.conda
       - conda: https://conda.anaconda.org/conda-forge/noarch/exceptiongroup-1.2.0-pyhd8ed1ab_2.conda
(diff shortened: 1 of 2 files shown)


JoerivanEngelen (approved):
Fantastic! This roughly reduces the duration of unittests from 35-45 minutes to 15 minutes, and examples from 30 minutes to 10 minutes!

PR #1050, opened 2024-05-22
Issue #1049 Restrict PATH to pixi and Modflow6

### LINKED ISSUE #1049: TeamCity: Restrict PATH variable on agents
We have some jobs randomly failing on certain agents, investigation of @Manangka showed that this is because of a global Anaconda install on the agents, which also comes with its own GDAL and NetCDF4 installations. Removing this installation from PATH removes the conflicting DLL loading.

As we cannot be certain what other colleagues outside our team install on the teamcity agents, it is better to run our jobs with a restricted PATH environmental variable.

JoerivanEngelen:
Fixes #1049

# Description
This PR restricts the PATH environmental variable for critical jobs on the TeamCity agent.

# Checklist

- [x] Links to correct issue
- [ ] Unit tests were added

### DIFF
--- a/.teamcity/Templates/ExamplesTemplate.kt
+++ b/.teamcity/Templates/ExamplesTemplate.kt
@@ -25,7 +25,7 @@ object ExamplesTemplate : Template({
             id = "Run_examples"
             workingDir = "imod-python"
             scriptContent = """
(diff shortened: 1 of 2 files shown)


Manangka:
Nice! Good to have full control over our dependencies again.

PR #1069, opened 2024-06-06
Issue #966 negative ibound

### LINKED ISSUE #966: Create converter iMOD5 to convert negative IBOUND values
In iMOD5, setting IBOUND < 0 meant heads were kept constant to their starting value. The work can be summarized in the following code:

```python
chd_heads = zeros_like(ibound) + shd_heads
chd_cells = ibound.where(ibound < 0)
chd_heads = chd_heads.where(chd_cells)
```

Where ``chd`` stands for ``ConstantHead`` and ``shd`` for ``InitialConcentration``. 

UPDATE
---------

- The above logic should be implemented in ``ConstantHead.from_imod5_ibound``
- ``ConstantHead.from_imod5_data`` should only create a package based on CHD packages present in the iMOD5 data. Try to keep this consistent with the equivalent methods in the Drainage and GHB package.

luitjansl:
Fixes #966

# Description
imports a constant head package using the ibound array and the chd-arrays in the imported project file data. 
These chd arrays are provided for the whole grid, but we assign constant head only in those cells that have ibound < 0.

# Checklist

- [X] Links to correct issue
- [X] Unit tests were added

### DIFF
--- a/imod/tests/test_mf6/test_mf6_chd.py
+++ b/imod/tests/test_mf6/test_mf6_chd.py
@@ -199,14 +199,13 @@ def test_write_concentration_period_data(head_fc, concentration_fc):
 def test_from_ibound(imod5_dataset):
     target_dis = StructuredDiscretization.from_imod5_data(imod5_dataset)
 
-    chd_layer_3 = imod.mf6.ConstantHead.from_imod5_ibound(
-        "chd-3",
+    chd_from_ibound = imod.mf6.ConstantHead.from_imod5_ibound(
         imod5_dataset,
         target_dis,
         regridder_types=None,
     )
-    errors = chd_layer_3._validate(
+    errors = chd_from_ibound._validate(
         imod.mf6.ConstantHead._write_schemata, idomain=target_dis["idomain"]
     )
-    errors.update(chd_layer_3._validate(imod.mf6.ConstantHead._init_schemata))
+    errors.update(chd_from_ibound._validate(imod.mf6.ConstantHead._init_schemata))
     assert len(errors) == 0
--- a/imod/tests/test_mf6/test_mf6_chd.py
+++ b/imod/tests/test_mf6/test_mf6_chd.py
@@ -205,6 +205,8 @@ def test_from_imod5(imod5_dataset, tmp_path):
         target_dis,
         regridder_types=None,
     )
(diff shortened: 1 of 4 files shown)


JoerivanEngelen (changes requested):
I think we have not properly refined this issue:

The logic to generate CHD from IBOUND=-1 looks fine, but iMOD5 can also have CHD packages itsself, for which  CHD packages need to be derived. On top of that, in case IBOUND=-1, another CHD package needs to be derived for this.

I think the ``from_imod5_data`` needs to derive a CHD package based on existing CHD data. Then in ``GroundwaterFlowModel.from_imod5_data``, a CHD package needs to be derived based on IBOUND=-1.

imod/tests/test_mf6/test_mf6_chd.py
```diff
@@ -192,3 +193,20 @@ def test_write_concentration_period_data(head_fc, concentration_fc):
...
                 data.count("2") == 1755
             )  # the number 2 is in the concentration data, and in the cell indices.
+
+
+@pytest.mark.usefixtures("imod5_dataset")
+def test_from_imod5(imod5_dataset):
+    target_dis = StructuredDiscretization.from_imod5_data(imod5_dataset)
+
+    chd_layer_3 = imod.mf6.ConstantHead.from_imod5_data(
+        "chd-3",
```
JoerivanEngelen, line 203: Hmmm, I remember now that the projectfile reader inferred separate CHD packages for each layer. I have to investigate again why that was the case, as I'm not really a fan of this.

luitjansl: the method "from_imod5_data" in the previous version of this PR is now renamed "from_imod5_ibound" and  only focusses on creating a chd package on the locations where ibound == -1. The starting head is used for the head in this chd package. Another issue (#1087) was created to import the other chd packages.

JoerivanEngelen:
I just had a chat with @FransRoelofsen and this confirmed our earlier suspicion (which renders my previous comment as incorrect):

iMOD5 uses continuous grids in the CHD package, which are assigned to cells with IBOUND = -1. If there are cells with IBOUND = -1, but no CHD package is defined, it takes the SHD package.

We can therefore make a single ``from_imod5_data`` method. That will work fine.

JoerivanEngelen:
Will be continued in https://github.com/Deltares/imod-python/pull/1088

PR #1080, opened 2024-06-18
Issue #1079 create python matrix

### LINKED ISSUE #1079: Use TeamCity matrix for the python pip builds

Manangka:
Fixes #1079

# Description
In the pipeline scripts we have 3 almost identical builds to test that pip install works with different python versions.
By using a build matrix we can test this in a single build allowing us to remove duplicate code.
Is also makes it easy to add/remove python versions to the pipeline script

image

image

# Checklist

- [x] Links to correct issue
- [ ] Unit tests were added


JoerivanEngelen (approved):
Looks good, I see the reporting from TeamCity reports that the build is stuck. I think this is resolved automatically when we merge this to master

PR #1082, opened 2024-06-19
Issue #1072 report faulty ipf during prj read

### LINKED ISSUE #1072: Projectfile: Report faulty IPF when throwing error
It is quite easy to have an IPF which is not parsable by iMOD Python.

However, the error thrown when reading an IPF reader only prints that there is an issue in the WEL package, but not which IPF. 

We should print/log the path or filename of the faulty IPF file.

JoerivanEngelen:
Fixes #1072 

# Description
This PR adds functionality to capture the path of a faulty IPF and IDF file in the error message.
For GEN files, this was already reported, but still added a unittest to verify.

# Checklist

- [x] Links to correct issue
- [x] Unit tests were added

### DIFF
--- a/docs/api/changelog.rst
+++ b/docs/api/changelog.rst
@@ -9,6 +9,11 @@ The format is based on `Keep a Changelog`_, and this project adheres to
 [Unreleased]
 ------------
 
+Fixed
+~~~~~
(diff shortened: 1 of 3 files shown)


luitjansl (approved):
looks good

PR #1086, opened 2024-06-24
issue #1083 duplicate wells 

### LINKED ISSUE #1083: on import of well ipf, deal with duplication of wells
sometimes multiple entries with the same ID can be found in an IPF file, 

- possibly with the same x and y coordinates
- sometimes with the same or different filter top/bottom
- sometimes with the same timeseries text file path

We distinguish exact spatial (not necessarily temporal)  duplication ( of x, y, filter_top, filter_bot, ID, but not necessarily rate )
and inexact  duplication where ID and possibly some but not all others of  x, y, filter_top, filter_bot.

For exact duplication we want to: 
-give a warning in the log file that this happened
-make the ID's unique by appending a suffix (_1, _2,..)
This should upon calling open_projectfile_data  (prj)

for inexact duplication: 
- we decide not to log this for now, it is a nice to have.

luitjansl:
Fixes #1083 

# Description
When importing wells, we now make sure that when the ipf file defines 2 or more wells with the same x, y , filter_top, filter_bottom and  id, then the ID of the second well is appended with "_1", and so on for successive findings of a well with the same characteristics (_2, _3 etc).

# Checklist

- [X] Links to correct issue
- [X] Unit tests were added


JoerivanEngelen (approved):
Looks good!

PR #1088, opened 2024-06-27
Issue #966 chd import or shd

### LINKED ISSUE #966: Create converter iMOD5 to convert negative IBOUND values

luitjansl:
Fixes #966

# Description
Imports chd packages present in imod5_data. These packages are imported one by one. If none are present, then a chd package is created at all locations where ibound == -1. In that case the starting head of the simulation is assigned as the chd value for each cell with ibound == -1.

# Checklist
- [X] Links to correct issue
- [X] Unit tests were added


imod/mf6/chd.py
```diff
@@ -141,3 +148,133 @@ def _validate(self, schemata, **kwargs):
...
+        target_discretization: StructuredDiscretization,
+        regridder_types: Optional[RegridMethodType] = None,
+    ) -> "ConstantHead":
+        target_idomain = target_discretization.dataset["idomain"]
+
+        # select locations where ibound < 0
+        head = head.where(ibound < 0)
+
+        # select locations where idomain > 0
+        head = head.where(target_idomain > 0)
```
JoerivanEngelen, line 263: ``head`` can be on a different grid than ibound and target_domain. Actually that is a very likely usecase, as the input for the CHD ``head`` usually comes from a different model simulation.

I therefore think it is safer to do this after regridding (also: This reminds me that we do not support not passing the ``regrid_context`` around yet, will make an issue for that).

imod/mf6/model_gwf.py
```diff
@@ -269,4 +269,16 @@ def from_imod5_data(
...
+        chd_keys = [key for key in imod5_keys if key[0:3] == "chd"]
+        if len(chd_keys) == 0:
+            result["chd_from_shd"] = ConstantHead.from_imod5_shd_data(
+                imod5_data, dis_pkg, regridder_types
+            )
+        else:
+            for chd_key in chd_keys:
+                result[chd_key] = ConstantHead.from_imod5_data(
+                    chd_key, imod5_data, dis_pkg, regridder_types
+                )
```
JoerivanEngelen, line 282: Our projectfile parser now specifies a seperate CHD package per layer. So this would result in multiple constant head packages being made, whereas the user probably expects 1. I'll create a separate issue for this; is out of scope for this PR.

JoerivanEngelen (approved):
Looks good, I like the seperate from_imod_5_shd method. I have one comment I'd like to be fixed: Doing the calls to ``where`` after regridding. The rest of these comments are just reminders for new issues, which I'll make now. Approving in advance.

imod/mf6/chd.py
luitjansl: done, now it regrids ibound as well, and it does the "where" statements with the regridded ibound.

PR #1089, opened 2024-06-27
Issue #598 layered hfb error

### LINKED ISSUE #598: Support multiple layers in the HFB layered HFB-classes
In #556 Layered HFB packages have been added. A user can define multiple linestring and per linestring a layer can be specified, However the layer specified for a linestring is applied to all linestrings

```
barrier_gdf = gpd.GeoDataFrame(
  geometry=
    [
      linestring1,
      linestring2
    ],
  data=
    {
      "hydraulic_characteristic": [1e-3, 1e-3,],
      "layer": [1, 2,]
    },
)


hfb = imod.mf6.LayeredHorizontalFlowBarrierHydraulicCharacteristic(barrier_gdf)
```

In the snippet above I would expect that linestring1 is applied to layer 1 and that linestring2 is applied to layer 2. However linestring1 is applied to layers 1 and 2 and the same applies for linestrin2

JoerivanEngelen:
Fixes #598

# Description
There was some confusion (also by me) about the use of the ``LayeredHorizontalFlowBarrier`` classes, so @HendrikKok explained me they were intended for single layers. I therefore changed the following to clarify:

- Rename to ``SingleLayerHorizontalFlowBarrier`` classes
- Throw ``ValidationError`` if multiple layers are added to dataset
- Added missing test for SingleLayerHorizontalFlowBarrier class

This means multiple SingleLayerHorizontalFlowBarrier need to be added to GroundwaterFlowModel object, whereas MODFLOW 6 only accepts one single HFB. So these have to be merged together. I think the best opportunity for this is to merge the low level ``Mf6HorizontalFlowBarrier`` objects.

# Checklist
- [x] Links to correct issue
- [x] Unit tests were added

### DIFF
--- a/imod/mf6/hfb.py
+++ b/imod/mf6/hfb.py
@@ -18,7 +18,7 @@ from imod.mf6.interfaces.ilinedatapackage import ILineDataPackage
 from imod.mf6.mf6_hfb_adapter import Mf6HorizontalFlowBarrier
 from imod.mf6.package import Package
 from imod.mf6.utilities.grid import broadcast_to_full_domain
-from imod.schemata import EmptyIndexesSchema
+from imod.schemata import EmptyIndexesSchema, NUniqueValueSchema
 from imod.typing import GridDataArray
 from imod.util.imports import MissingOptionalModule
 
@@ -283,7 +283,7 @@ class HorizontalFlowBarrierBase(BoundaryCondition, ILineDataPackage):
 
     _period_data = ()
     _init_schemata = {}
-    _write_schemata = {"geometry": [EmptyIndexesSchema()]}
+    _write_schemata = {"geometry": [EmptyIndexesSchema()], "layer": [NUniqueValueSchema(1)]}
 
     def __init__(
         self,
@@ -342,7 +342,6 @@ class HorizontalFlowBarrierBase(BoundaryCondition, ILineDataPackage):
         top: GridDataArray,
         bottom: GridDataArray,
         k: GridDataArray,
-        validate: bool = False,
     ) -> Mf6HorizontalFlowBarrier:
         """
         Write package to Modflow 6 package.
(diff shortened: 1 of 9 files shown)


imod/schemata.py
```diff
@@ -503,6 +503,21 @@ def validate(self, obj: GridDataArray, **kwargs) -> None:
             )
 
 
+class NUniqueValueSchema(BaseSchema):
+    """
+    Schema to validate if amount of unique values is correct.
```
luitjansl, line 508: It appears that the validate method reports an error when the number of unique values exceeds the indicated amount ("length"), but not when it has less unique values than the indicated amount. Please modify the docstring to stress that.

imod/schemata.py
```diff
@@ -503,6 +503,21 @@ def validate(self, obj: GridDataArray, **kwargs) -> None:
...
+    """
+
+    def __init__(self, length: int):
```
luitjansl, line 511: length ->max_unique_values

imod/mf6/hfb.py
```diff
@@ -1105,4 +1120,4 @@ def from_imod5_dataset(cls, imod5_data: dict[str, dict[str, GridDataArray]]):
             dataframe_ls.append(geometry_layer)
         compound_dataframe = pd.concat(dataframe_ls, ignore_index=True)
 
-        return LayeredHorizontalFlowBarrierResistance(compound_dataframe)
+        return cls(compound_dataframe)
```
luitjansl, line 1123: there is also a "LayeredHorizontalFlowBarrierHydraulicCharacteristic". 
Should the changes in this PR also be applied there?

luitjansl (commented):
looks good, some comments

imod/mf6/hfb.py
JoerivanEngelen: I think not. The hydraulic characteristic is a MODFLOW 6 concept: it is the inverse of the resistance. In the end, both the ``Resistance`` as well as the ``HydraulicCharacteristic`` variant are converted to a hydraulic characteristic for the Modflow6 HFB package. iMOD5 only allows you to specify a hfb with a resistance, so I don't see a real use case for adding a ``from_imod5_data`` method to the HydraulicCharacteristic class.

imod/schemata.py
JoerivanEngelen: Good one

imod/schemata.py
JoerivanEngelen: Good one, will improve!

luitjansl (approved):

PR #1104, opened 2024-07-08
added logging and error messages for well import when

### LINKED ISSUE #1095: Enforce autoplacement of wells in the data loaded with .open_projectfile_data
Our ``Well`` package doesn't support a layer specified by the user. iMOD5 does support this.
To fix this, we might require something like a ``LayeredWell`` object, congruent with the HFB package.

Requirements
---------------

- If necessary, force autoplacement of wells, for example by setting layer number to 0 in the dictionary loaded with ``open_projectfile_data``
- Log warning that we are not supporting placing wells to a specified layer
- If ``filter_top`` and ``filter_bottom`` are not present in the data, throw error upon calling ``from_imod5_data``.

luitjansl:
Fixes #1095 

# Description
This PR adds logging to well import if the well was assigned to a layer. This layer will not be used and the user is informed of that in the log. 
If the filter_top and filter_bottom properties are not present in the imported dataset, both a log message and an exception is generated, as we can't import the well with this information missing

# Checklist

- [X] Links to correct issue
- [X] Unit tests were added

### DIFF
--- a/imod/mf6/wel.py
+++ b/imod/mf6/wel.py
@@ -653,6 +653,31 @@ class Well(BoundaryCondition, IPointDataPackage):
         minimum_k: float = 0.1,
         minimum_thickness: float = 1.0,
     ) -> "Well":
+        if "layer" in imod5_data[key].keys():
+            if imod5_data[key]["layer"] != 0:
+                log_msg = textwrap.dedent(
+                    f"""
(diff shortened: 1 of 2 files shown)


JoerivanEngelen (approved):
Looks good!

Issue #1116, opened 2024-07-23
Dealing with WEL IPFs from a project file in open_projectfile_data


Huite:
There are a number of issues with the current implementation of `open_projectfile_data`:

It loads identical dataframes over and over, because it treats every row of every WEL block separately. This is very inefficient.
Secondly, it doesn't reproduce the behavior of iMOD5 unless very specific conditions are met. For simplicity, I'll categorize the IPFs into two categories: "simple" when it's just a single file with rows of (x, y, rate), and "associated" when it contains references to the external time series (txt) files.

The key part to understand here is that iMOD5 clears **all** wells whenever it reads a new block. Due to the way that we work, by essentially creating discharge time series of each well first, we can only reproduce it by introducing a zero rate value at the the next stress period. Essentially, whenever we define some input, we could duplicate the stress period data, set its rate value to 0.0, and append it -- I'm not sure this works though, it depends on how the dataframes are combined afterwards.

Anyway, to give some example, a single definition:

```
0001,(WEL),1
1900-01-01
001,002
 1,2, 001, 1.0, 0.0, -999.9900 ,"simple.ipf"
 1,2, 001, 1.0, 0.0, -999.9900 ,"associated.ipf"
```

Here, the values of simple are simply true for the all subsequent stress periods.
For associated, however, iMOD will read the associated text files **every** stress period and determine the rate value.

```
0001,(WEL),1
1900-01-01
001,002
 1,2, 001, 1.0, 0.0, -999.9900 ,"simple.ipf"
 1,2, 001, 1.0, 0.0, -999.9900 ,"associated.ipf"
1900-01-02
001,001
 1,2, 001, 1.0, 0.0, -999.9900 ,"simple.ipf"
```

Note: This will turn off the associated wells at the second stress period!
Essentially, there's a duplication of time series data: once in project file, and then again in the associated text files.

Alternatively:

```
0001,(WEL),1
1900-01-01
001,001
 1,2, 001, 1.0, 0.0, -999.9900 ,"simple.ipf"
1900-01-02
001,001
 1,2, 001, 1.0, 0.0, -999.9900 ,"associated.ipf"
```

In this case, associated is only turned on in the second stress period. Unfortunately, for us, this would mean that we would have to cut off any data in the associated files that exists prior to 1900-01-02. I think this is quite undesirable: it forces time selection data into the data loading routines, rather complicating it.

[...]

PR #1117, opened 2024-07-24
Start addressing #1116

Huite:
This is a start of addressing the issues described in #1116.

I haven't run any of this code, but it does contain most of the ideas. In my mind, this:

* has reasonable demands of the project file;
* gives a relatively clear error message if those demands are not met;
* returns a reasonable result for further processing.

The last point is the most contentious. You have logic for converting the list of dicts with dataframes into a dataset.
What this PR does address is the fact that for the simple IDFs, they have to be "turned off" every next stress period definition: essentially, we are turning the project file + simple IDFs into a time series for each well point too.

However, this requires inserting starting and end times into the dataframes. That could be done here, but it may be easier to do in the subsequent processing step. I find that difficult to judge.
Note that the associated IPFs already have a time column.

### DIFF
--- a/imod/formats/prj/prj.py
+++ b/imod/formats/prj/prj.py
@@ -3,12 +3,13 @@ Utilities for parsing a project file.
 """
 
 import shlex
+import textwrap
 from collections import defaultdict
 from datetime import datetime
 from itertools import chain
 from os import PathLike
 from pathlib import Path
-from typing import Any, Dict, List, Sequence, Tuple, Union
+from typing import Any, Dict, List, NamedTuple, Sequence, Tuple, Union
 
 import numpy as np
 import pandas as pd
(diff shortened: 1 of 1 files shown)


Huite:
I've currently just written the helper functions in the body of the main function. Given the complexity here, it's probably better to move it into a separate module. There's already a prj folder, after all. The only thing to keep in mind is that _try_read_with_func should then be split off into another module to avoid circular imports.

Huite:
I think I've just thought of another complication: this currently uses the associated IPF entries of the first stress period block to define them. However, it's technically possible to use the same associated IPF twice, but with a different layer entry. This means the current check is inadequate. So technically, I guess any unique (layer, associated IPF) is fine.

However, I think it's more reasonable to disallow assigning an associated IPF to more than one layer. It's pretty weird, and seems more likely to be the result of user error.

JoerivanEngelen:
Fixed in https://github.com/Deltares/imod-python/pull/1142

PR #1118, opened 2024-07-24
Issue #1092 merge chd

### LINKED ISSUE #1092: ``GroundwaterFlowModel.from_imod5_data`` should merge CHD packages into 1
The ``open_projectfile_data`` creates separate CHD packages per layer. 

The context for this is that iMOD5 supports specifying multiple systems for a single package. This means you can specify multiple river stages and conductances to a single cell, which is useful for example in Dutch polder systems, where there are a lot of small surface water networks set to different stages in a single model cell. The projectfile doesn't have a way to discern which layer data belongs to which system for boundary condition entries. It is hard to figure out programmatically which layers form 1 system. Users do this by manually inspecting the paths to their IDFs which they hopefully provided with a good naming. As there is no strict naming convention for these systems in the paths, we should make separate packages for each layer. Users can manually merge systems together in a script if they want to.

The CHD package, however, is an exception, as it sets heads directly in the groundwater model (so a Dirichlet instead of Robin boundary condition). This means it doesn't make sense to specify multiple heads in a single cell, rendering separate CHD systems useless. We can therefore safely assume that the different CHD entries are different layers in a single system, and combine them into one single CHD package.

Context
---------
- Grid based boundary conditions are always defined for all timesteps. 

Requirements
---------------
- Create separate function, something like, ``concat_layered_chd_packages``, which stacks the separate packages into one. Keys of these packages are "chd-1", "chd-2", where the number indicates layer number.
- We can use [xr.concat](https://docs.xarray.dev/en/stable/user-guide/combining.html#concatenate) for this. To be on the safe side, set ``compat`` setting to ``"equals"``
- Call function in ``GroundwaterFlowModel.from_imod5_data

Nice to have
--------------
- Seperate logic to stack layers in a helper function (we probably already have logic for this). Function is focused on stacking layers.

luitjansl:
Fixes #1092 

# Description
On import, chd packages that were split over layers are merged into a single chd package if the names of these packages are 
"chd-x" where x is a layer number. 
A function for this was added to mf6/utilities and not in the chd package to keep it hidden from users. 

# Checklist

- [X] Links to correct issue
- [X] Unit tests were added

### DIFF
--- a/imod/mf6/utilities/chd_concat.py
+++ b/imod/mf6/utilities/chd_concat.py
@@ -1,0 +1,48 @@
+from typing import Optional
+
+import xarray as xr
+
+from imod.mf6.chd import ConstantHead
+
+
+def concat_layered_chd_packages(
+    name: str,
+    dict_packages: dict[str, ConstantHead],
+    remove_merged_packages: bool = True,
+) -> Optional[ConstantHead]:
+    """
+
+    Parameters
+    ----------
+    name: str
+        The name of the package that was split over layers.
+        If they are called "chd-1" and so on then set name to "chd
+    dict_packages: dict[str, ConstantHead]
+        dictionary  with package names as key and the packages as values
+    remove_merged_packages: bool = True
+        set to True to remove merged packages from dict_packages
+
+    This function merges chd-packages whose name starts with "name" into a a
+    single chd package. This is aimed at chd packages that are split over
+    layers- so we would have chd-1, chd-2 and so on and these packages would
+    define a chd package for layer 1, 2 and so on. This function merges them
+    into a single chd package. If remove_merged_packages is True, then the
+    packages that are concatenated are removed from the input dictionary, so
+    that this on output only contains the packages that were not merged.
+    """
+
+    candidate_keys = [k for k in dict_packages.keys() if name in k]
+    if len(candidate_keys) == 0:
+        return None
+
+    dataset_list = []
+    for key in candidate_keys:
+        pack = dict_packages[key]
+        dataset_list.append(pack.dataset)
+        if remove_merged_packages:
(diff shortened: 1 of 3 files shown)


imod/mf6/utilities/chd_concat.py
```diff
@@ -0,0 +1,48 @@
...
+    dict_packages: dict[str, ConstantHead],
+    remove_merged_packages: bool = True,
+) -> Optional[ConstantHead]:
+    """
+
+    Parameters
+    ----------
+    name: str
+        The name of the package that was split over layers.
+        If they are called "chd-1" and so on then set name to "chd
```
JoerivanEngelen, line 19:
```suggestion
        If they are called "chd-1" and so on then set name to "chd"
```

imod/mf6/utilities/chd_concat.py
```diff
@@ -0,0 +1,48 @@
...
+    dict_packages: dict[str, ConstantHead]
+        dictionary  with package names as key and the packages as values
+    remove_merged_packages: bool = True
```
JoerivanEngelen, line 22: I have some doubts about the ``remove_merged_packages`` option. 

I think you implemented it to avoid double memory usage in the ``GroundwaterFlow.from_imod5_data`` method. However, this happens also in ``concat_layered_chd_packages`` anyway (dataset per layer in list, and concatenated dataset). The duplicate memory use now avoided in ``GroundwaterFlow.from_imod5_data`` is of little use as ``concat_layered_chd_packages`` is called at the end.

JoerivanEngelen (commented):
Looks good. I only have some doubts about the ``remove_merged_packages`` option, as I see no real use for it, see explanation below. Could you elaborate the use case for it?

PR #1121, opened 2024-07-26
Issue #1096 layered wells

### LINKED ISSUE #1096: Support user-specified layers for Wells
For iMOD5, and also some users might like this, we require that wells can be placed in a user-defined layer.
The ``WellDisVertices`` and ``WellDisStructured`` objects allow us to do this, but these are phased out, as they require cells to be defined as layer, row, column, whereas we require them to be defined as layer, y, x. The ``Well`` class only supports placing wells with [filt_top, filt_bot], y, x, not with layer. 

The pragmatic solution with the current code base would be looking up the ``filter_top`` and ``filter_bot`` in the top and bottom grids, based on the layer, y, x well locations. This has the following disadvantages:

- Logic is weird to user: The user has specified wells to a layer in their projectfile, but then only sees a filter_top and filter_bot in the dictionary they get
- The dataflow becomes more complex, as an addition data transformation is added, requiring extra logic to maintain.
- As model top and bottom are regridded, a well might suddenly be allocated to more layers than specified in the projectfile. This differs from iMOD5 behavior

The alternative is creating a separate `LayeredWell` object or something like that. Congruent with the ``LayeredHorizontalFlowBarrier`` classes (renamed to ``SingleLayerHorizontalFlowBarrier`` class soon). We have to work out if the ``LayeredWell`` class will support multiple layers or not. This has the disadvantage that:

- The architecture becomes more complex: another object is added, requiring maintenance and tests.
- The filter length can change unknowingly as model tops and bottoms are changed when regridding. This also happens in iMOD5, but might be unknown to a user.

The last points of each solution are opposites, and it depends on a user's expectations what is preferred. Either we should keep things as much to iMOD5 as possible, or require some flexibility of users to adapt to the iMOD Python philosophy.

JoerivanEngelen:
Fixes #1096

# Description

This changes:

- a ``LayeredWell`` object
- separate relevant code from ``Well`` class into more generic abstract base class: ``GridAgnosticWell``
- seperate test cases for ``to_mf6_pkg``, to also add test cases for ``LayeredWell``

I haven't added a ``from_imod5_data`` method, which would require some extra work. This still needs to be done, in a follow-up PR and issue. I for now refrained from doing this to make the changeset more accessible to review.

# Checklist

- [x] Links to correct issue
- [x] Unit tests were added

### DIFF
--- a/docs/api/changelog.rst
+++ b/docs/api/changelog.rst
@@ -40,6 +40,8 @@ Added
   :class:`imod.mf6.HorizontalFlowBarrierResistance`,
   :class:`imod.mf6.HorizontalFlowBarrierMultiplier`,
   :class:`imod.mf6.HorizontalFlowBarrierHydraulicCharacteristic`.
+- :class:`imod.mf6.LayeredWell` to specify wells directly to layers instead
+  assigning them with filter depths.
 
 
 [Unreleased]
--- a/imod/mf6/__init__.py
+++ b/imod/mf6/__init__.py
@@ -49,5 +49,5 @@ from imod.mf6.sto import SpecificStorage, Storage, StorageCoefficient
 from imod.mf6.timedis import TimeDiscretization
 from imod.mf6.utilities.regrid import RegridderType, RegridderWeightsCache
 from imod.mf6.uzf import UnsaturatedZoneFlow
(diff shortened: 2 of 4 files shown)


JoerivanEngelen:
Merged in https://github.com/Deltares/imod-python/pull/1123

PR #1127, opened 2024-07-29
Issue #1120 import steady state lhm fixes

### LINKED ISSUE #1120: Try reading and writing the LHM steady-state
@HendrikKok tried to read and convert the LHM as created by Peter. He succeeded, but it required some manual fixes. I think it would be worthwhile to do that exercise again and note down all the issues we encounter. Furthermore we can do some profiling on this if we deem this conversion too slow (e.g. slower than iMOD5).

Context
=====
This LHM was already converted from quasi-3D to fully 3D, which is a requirement.

Known issues
=========
@HendrikKok already encountered the following issues during conversion:

iMOD Python
---------------
- Wells will not work yet until #1116 and #1096 are fixed, remove those from the projectfile.
- There is no storage package, this will make iMOD Python throw an error. Hendrik added it manually. We could turn off the validation for this. 
- Apparently we require a recharge package somewhere
- He was unable to dump a model with the HFB package, this is worthy of a separate issue.

The model
------------
There are also some issues with the model:

- I recall there was some error in a path. This will be thrown when calling ``open_projectfile_data``. Contact Hendrik if we cannot find the cause.
- The river packages will throw validation errors; there are still some mistakes in it apparently.

luitjansl:
Fixes #1120

# Description
this PR contains fixes made to be able to import a version of the LHM that is steady state.
**Probably not all these changes should be merged.  The review should reveal what to keep and what not.** 

Changes:
-sto package can be absent in the imod5 project file. If that is the case, now a default sto package is made which imposes steady state. 
-rch package can now be absent in the project file. Then no recharge is present in the simulation. 
- allocation option and distribution option for river imports were  hard coded to different defaults . The old defaults gave an error, which could be a new issue in itself.
-also for GHB a new default was set, because the old one gave an error. This could be a new issue in itself.

# Checklist

- [X] Links to correct issue
- [X] Unit tests were added


JoerivanEngelen (commented):
Looks good, except the fixes you make in the River.from_imod5_data. I think these can be moved to public API utility functions. Furthermore I have an optimization suggestion to ``mask_arrays``.

imod/mf6/riv.py
```diff
@@ -331,13 +332,33 @@ def from_imod5_data(
...
+            regridded_package_data["bottom_elevation"],
+            target_bottom_subset,
+        )
+        regridded_package_data = mask_arrays(regridded_package_data)
+        regridded_package_data["stage"] = xr.where(
+            regridded_package_data["stage"]
+            > regridded_package_data["bottom_elevation"],
+            regridded_package_data["stage"],
+            regridded_package_data["bottom_elevation"],
+        )
```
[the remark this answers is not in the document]
luitjansl: made issue #1131 for this

imod/mf6/riv.py
```diff
@@ -335,9 +336,14 @@ def from_imod5_data(
         optional_river_package: Optional[River] = None
         optional_drainage_package: Optional[Drainage] = None
         # create a drainage package with the conductance we computed from the infiltration factor
+        drainage_arrays = {
+            "stage": regridded_package_data["stage"],
+            "conductance": drain_conductance,
+        }
+        drainage_arrays = mask_arrays(drainage_arrays)
```
JoerivanEngelen, line 343: This is still some leftover behind-the-scenes cleaning, which we need to avoid. Could you revert this as well?

luitjansl: done

PR #1140, opened 2024-08-05
Issue #1139 apply factor and addition

### LINKED ISSUE #1139: speed up chd import for Mipwa regional model
The import of time-dependent chd data seems to be a time-sink on importing the Mipwa regional model. Profiling showed that much of the time was spent in apply_factor_and_addition which applies the user-specified factor and addition to the ipf data associated to the chd in a particular timestep.

luitjansl:
Fixes #1139 

# Description
Loading project files turned out to be a time-sink for project file imports- especially when loading prj-file data with transient arrays such as chd-levels per timestep. 
These arrays were loaded in a function called "apply_factor_and_addition" which accounted for the fact that the user can specify a factor to multiply each array and a constant to add to it. These 2 (factor and addition) can be different for every timestep. 
Now they are applied at an earlier stage, when loading the ipf data. 

No new tests were added, because the tests in  test_import_prj.py already test the application of factors and additions. (this PR is for improving performance and not changing behavior)

# Checklist

- [ ] Links to correct issue
- [ ] Unit tests were added

### DIFF
--- a/imod/formats/prj/prj.py
+++ b/imod/formats/prj/prj.py
@@ -522,21 +522,24 @@ def _try_read_with_func(func, path, *args, **kwargs):
         raise type(e)(f"{e}. Error thrown while opening file: {path}")
 
 
-def _create_arithmetic_dataarray(
+def _get_array_transformation_parameters(
     headers: List[Dict[str, Any]], key: str, dim: str
 ) -> xr.DataArray:
-    return xr.DataArray(
-        data=[header[key] for header in headers],
-        dims=(dim,),
-        coords={dim: [header[dim] for header in headers]},
-    )
+    if dim in headers[0].keys():
+        return xr.DataArray(
+            data=[header[key] for header in headers],
+            dims=(dim,),
+            coords={dim: [header[dim] for header in headers]},
+        )
+    else:
+        return headers[0][key]
 
 
 def _create_datarray_from_paths(
     paths: List[str], headers: List[Dict[str, Any]], dim: str
 ) -> xr.DataArray:
-    factor = _create_arithmetic_dataarray(headers, "factor", dim)
-    addition = _create_arithmetic_dataarray(headers, "addition", dim)
+    factor = _get_array_transformation_parameters(headers, "factor", dim)
+    addition = _get_array_transformation_parameters(headers, "addition", dim)
     da = _try_read_with_func(
         imod.formats.array_io.reading._load,
         paths,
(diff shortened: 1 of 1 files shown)


imod/formats/prj/prj.py
```diff
@@ -547,6 +547,15 @@ def _create_dataarray_from_paths(
         _read=imod.idf._read,
         headers=headers,
     )
+
+    # remove coordinates in factor and addition but not in da
+    if type(factor) is xr.DataArray:
+        for d in factor.dims:
+            if d not in da.dims:
+                if len(factor.coords[d]) == 1:
+                    factor = factor.isel({d: 0}).drop(d)
```
Huite, line 556: `.isel({d: 0}, drop=True)` is probably what you're after?

imod/formats/prj/prj.py
```diff
@@ -547,6 +547,15 @@ def _create_dataarray_from_paths(
         _read=imod.idf._read,
         headers=headers,
     )
+
+    # remove coordinates in factor and addition but not in da
+    if type(factor) is xr.DataArray:
+        for d in factor.dims:
+            if d not in da.dims:
```
Huite, line 554: You can probably do this at once with a set:

```python
missing_dims = set(factors.dims) - da.dims
if missing_dims:
     factor = factor.isel({d: 0 for d in missing_dims}, drop=True)
```

imod/formats/prj/prj.py
```diff
@@ -547,6 +547,15 @@ def _create_dataarray_from_paths(
         _read=imod.idf._read,
         headers=headers,
     )
+
+    # remove coordinates in factor and addition but not in da
+    if type(factor) is xr.DataArray:
```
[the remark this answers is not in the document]
luitjansl: done

imod/formats/prj/prj.py
luitjansl: done

imod/formats/prj/prj.py
luitjansl: done

imod/formats/prj/prj.py
```diff
@@ -522,30 +522,63 @@ def _try_read_with_func(func, path, *args, **kwargs):
         raise type(e)(f"{e}. Error thrown while opening file: {path}")
 
 
-def _create_datarray_from_paths(paths: List[str], headers: List[Dict[str, Any]]):
+def _get_array_transformation_parameters(
+    headers: List[Dict[str, Any]], key: str, dim: str
+) -> xr.DataArray:
+    if dim in headers[0].keys():
```
JoerivanEngelen, line 528: Maybe add a short docstring what this helper function does?

imod/formats/prj/prj.py
```diff
@@ -522,30 +522,63 @@ def _try_read_with_func(func, path, *args, **kwargs):
...
     da = _try_read_with_func(
         imod.formats.array_io.reading._load,
         paths,
         use_cftime=False,
         _read=imod.idf._read,
         headers=headers,
     )
-    return da
 
+    # remove coordinates in factor and addition but not in da
```
JoerivanEngelen, line 551: I think a more specific comment here is:

_"Ensure factor and addition do not have more dimensions than da"_

JoerivanEngelen (approved):
Looks good, 2 minor comments/suggestions. Approving in advance

imod/formats/prj/prj.py
luitjansl: done

imod/formats/prj/prj.py
luitjansl: done

PR #1157, opened 2024-08-16
Issue #1132 performance hfb

### LINKED ISSUE #1132: Poor performance writing layered HFBs
Some context
---------------
- A ``LayeredHorizontalFlowBarrier`` was introduced, renamed now to ``SingleLayerHorizontalFlowBarrier`` as it only worked for single layers. 
- In the LHM, and other regional models, multiple layers can be assigned the same set of lines (same GEN file).

Issue
-----

User testing from @luitjansl showed that writing the steady-state LHM takes ~2 minutes without HFBs and ~25 minutes with HFBs, which is quite killing for user's productivity.  The reason is that a lot of calls are made to ``xu.snap_to_edges``. 

This is because:

- A call needs to be made for every linestring, as there is no support for summation of resistances in xugrid
- A call needs to be made for every single layer, as we are dealing with SingleLayerHorizontalFlowBarriers

Potential solutions
--------------------

I see three potential optimizations, in order of perceived difficulty to implement, low to high:

1. Support in iMOD Python that lines assigned to multiple layers are snapped to cell edges only once. Either by creating a LayeredHorizontalFlowBarrier and combine all hfb systems that point to the same GENfile into it. Or by caching snapped edges for each file.
2. Add support in xugrid to snap and aggregate to edges.
3. Create separate optimized snapping functions for structured grids

Final note
-----------

This is not so much of a problem for the HorizontalFlowBarrier with a ztop and zbottom. There snapping only is done once for all layers automatically.

JoerivanEngelen:
Fixes #1132 

# Description
This PR adds the following:

- Increase performance for writing the HFB files, by using new xugrid create_snap_to_grid_dataframe.
- Fixes error when multiple barriers in a single dataset would be snapped to same edge. These lines are now aggregated.
- Add some basic infrastructure for user acceptance test for writing LHM, the path to the LHM projectfile should be set in a .env as LHM_PRJ. This can later be extended to TeamCity.

This reduces the time spent writing the HFB packages from 34 minutes to 12.5 minutes on my local machine. I've profiled the causes of the remaining bottlenecks to writing. I'll post a follow-up issue for that.

# Checklist

- [x] Links to correct issue
- [x] Unit tests were added

### DIFF
--- a/imod/mf6/hfb.py
+++ b/imod/mf6/hfb.py
@@ -391,6 +391,13 @@ def _snap_to_grid_and_aggregate(
         Grid to snap lines to
     vardict_agg: dict
         Mapping of variable name to aggregation method
+    
+    Returns
+    -------
+    snapping_dataset: xugrid.UgridDataset
+        Dataset with all variables snapped and aggregated to cell edges
+    edge_index: numpy.array
+        1D array with indices of cell edges that lines were snapped to
     """
     snapping_df = xu.create_snap_to_grid_dataframe(
         barrier_dataframe, grid2d, max_snap_distance=0.5
--- a/imod/mf6/hfb.py
+++ b/imod/mf6/hfb.py
@@ -28,7 +28,7 @@ from imod.mf6.utilities.hfb import (
 )
 from imod.schemata import EmptyIndexesSchema, MaxNUniqueValuesSchema
 from imod.typing import GeoDataFrameType, GridDataArray, LineStringType
-from imod.typing.grid import enforce_uda
+from imod.typing.grid import as_ugrid_dataarray
 from imod.util.imports import MissingOptionalModule
 
 if TYPE_CHECKING:
@@ -514,7 +514,7 @@ class HorizontalFlowBarrierBase(BoundaryCondition, ILineDataPackage):
         k = idomain * k
         # Enforce unstructured
         unstructured_grid, top, bottom, k = (
-            enforce_uda(grid) for grid in [idomain, top, bottom, k]
+            as_ugrid_dataarray(grid) for grid in [idomain, top, bottom, k]
         )
         snapped_dataset, edge_index = self._snap_to_grid(idomain)
         edge_index = self.__remove_invalid_edges(unstructured_grid, edge_index)
@@ -810,7 +810,7 @@ class HorizontalFlowBarrierBase(BoundaryCondition, ILineDataPackage):
         if has_layer:
             vardict_agg["layer"] = "first"
         # Create grid from structured
-        grid2d = enforce_uda(idomain.sel(layer=1)).grid
+        grid2d = as_ugrid_dataarray(idomain.sel(layer=1)).grid
 
         return _snap_to_grid_and_aggregate(barrier_dataframe, grid2d, vardict_agg)
 
--- a/imod/mf6/hfb.py
+++ b/imod/mf6/hfb.py
@@ -377,7 +377,7 @@ def _prepare_barrier_dataset_for_mf6_adapter(dataset: xr.Dataset) -> xr.Dataset:
 
 
 def _snap_to_grid_and_aggregate(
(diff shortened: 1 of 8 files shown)


imod/mf6/hfb.py
```diff
@@ -374,6 +376,45 @@ def _prepare_barrier_dataset_for_mf6_adapter(dataset: xr.Dataset) -> xr.Dataset:
     return dataset
 
 
+def _snap_to_grid_and_aggregate(
```
luitjansl, line 379: this function may deserve its own test

imod/mf6/hfb.py
```diff
@@ -374,6 +376,45 @@ def _prepare_barrier_dataset_for_mf6_adapter(dataset: xr.Dataset) -> xr.Dataset:
...
+    barrier_dataframe: GeoDataFrameType, grid2d: xu.Ugrid2d, vardict_agg: dict[str, str]
+) -> tuple[xu.UgridDataset, npt.NDArray]:
+    """
+    Snap barrier dataframe to grid and aggregate multiple lines with a list of
+    methods per variable.
```
luitjansl, line 384: The description doesn't tell what  it returns

pyproject.toml
```diff
@@ -157,6 +157,7 @@ ignore_missing_imports = true
 [tool.pytest.ini_options]
 markers = [
     "example: marks test as example (deselect with '-m \"not example\"')",
+    "user_acceptance: marks user acceptance tests (deselect with '-m \"not user_acceptance\"')",
```
luitjansl, line 160: what's the default? will running the tests from vscode run the user acceptance tests as well?

imod/tests/test_mf6/test_mf6_LHM.py
```diff
@@ -0,0 +1,94 @@
...
+        mf6_hfb_ls = []
+        for key, pkg in model.items():
+            if issubclass(type(pkg), imod.mf6.HorizontalFlowBarrierBase):
+                mf6_hfb_ls.append(pkg)
+            pkg.dataset.load()
+
+        top, bottom, idomain = model._Modflow6Model__get_domain_geometry()
+        k = model._Modflow6Model__get_k()
+
+        mf6_hfb = merge_hfb_packages(mf6_hfb_ls, idomain, top, bottom, k)
```
luitjansl, line 86: can merge_hfb_package be tested in a reguar unit test? If it is in the acceptance tests, it may not be tested very often.

imod/typing/grid.py
```diff
@@ -431,3 +433,20 @@ def is_transient_data_grid(
         if len(grid["time"]) > 1:
             return True
     return False
+
+
+@typedispatch
+def enforce_uda(grid: xr.DataArray) -> xu.UgridDataArray:
```
luitjansl, line 439: rename to as_xugrid_dataarray?

luitjansl (commented):
looks good!

pyproject.toml
JoerivanEngelen: Calling all tests from vscode on the left hand side will indeed run the user acceptance tests. Just as this will run the examples.

However, calling ``pixi run unittests`` will ignore the user acceptance tests, as well as ignore the examples, as well as run the tests in parallel. So think running all tests with ``pixi run unittests`` is adviceable, and I'd only use the VSCode test interface for subsets of unittests.

imod/tests/test_mf6/test_mf6_LHM.py
JoerivanEngelen: ``merge_hfb_packages`` has separate unittests in ``test_mf6hfb.py``, so that is tested regularly

imod/typing/grid.py
JoerivanEngelen: Good suggestion, was struggling a bit with naming here!

PR #1159, opened 2024-08-16
Issue #1158 remaining hfb bottlenecks

### LINKED ISSUE #1158: Bottlenecks writing HFBs LHM
https://github.com/Deltares/imod-python/pull/1157 considerably improves the speed with which the LHM is written: From 34 minutes to 12.5 minutes. This is still slower than expected however. Based on profiling, I've identified 3 main bottlenecks:

- 4.5 minutes: ``xu.DataArray.from_structured`` is called 70 times. This is quite a computationally intensive operation. It can probably be reworked to be called only once.
- 3 minutes are spent in ``DTypeSchema.validate``. This make as call to the ``scalar_None`` function, which is supposed to shortcut if len(da.shape)  != 0, but because I used the ``&`` operator accidentily instead of ``and``, this is not happening.
- 3 minutes: Spent in ``mask_all_models``. This is called because the automatic regridding might introduce inconsistencies in where idomain is located and where data is specified.  I think we can remove this function, because:
     1. in this case no regridding takes place and the masking is unnecessary, besides I see that ``_regrid_like(model: IModel, *args)`` also does masking?
     2. automatic masking can hide issues already present in the original iMOD5 model. 
     3. Regridding might introduce other issues as well: For example, model tops and bottoms might be regridded in such a manner that the bottom exceeds top. Automatically cleaning up all these potentially introduced artifacts is excessive. Probably better is to create cleanup utilities and guide the user to these in the ValidationError messages. 

So tasks:

- [ ] Ensure  ``xu.DataArray.from_structured`` is only called once
- [ ] Make sure ``scalar_None`` shortcuts appropriately
- [ ] Remove call to mask_all_models

JoerivanEngelen:
Fixes #1158 

# Description
Fix remaining HFB bottlenecks. This reduces writing the HFB package from the LHM from 12.5 minutes to 2 minutes.

- Cache results ``xu.Ugrid2d.from_structured``, as this is a costly operation
- Use ``and`` instead of ``&`` operator in the ``scalar_None`` function, to enable shortcutting.
- Remove ``mask_all_packages`` function in ``from_imod5_data``, call in tests.
- Create separate pixi task for slow unittests, where just in time compilation is enabled. ``pixi run unittests`` still runs all unittests, by starting two pixi tasks ``unittests_njit`` & ``unittests_jit``.

# Checklist
- [x] Links to correct issue
- [x] Unit tests were added


imod/mf6/simulation.py
```diff
@@ -1378,8 +1378,5 @@ def from_imod5_data(
         )
         simulation["ims"] = solution
 
-        # cleanup packages for validation
-        idomain = groundwaterFlowModel.domain
-        simulation.mask_all_models(idomain)
```
luitjansl, line 1383: there are a few examples on regridding. Even if  they don't need this masking, it may be helpful to modify them to do it anyway to make the reader aware  how to mask all the regridded packages

luitjansl (approved):
looks good.1 comment, approving in advance

imod/mf6/simulation.py
JoerivanEngelen: The hondsrug example is the example where we regrid a simulation. In this example data is already made consistent, and regridding doesn't introduce any problems. I defer from adding unnecessary calls to the example, as it might lead users to blindly call computationally intensive functions. However, I want to create an example how to import a (problematic) iMOD5 model, where some cleanup needs to be done (for which cleanup utilities have to be made). We could introduce that here. I'll create a separate issue for that.

JoerivanEngelen: See issue here https://github.com/Deltares/imod-python/issues/1164

PR #1160, opened 2024-08-18
[TEAMCITY] Update project dependencies

deltares-service-account:
Teamcity automatically updated the dependencies defined the pixi.toml file. Please verify that all tests succeed before merging


Manangka:
@JoerivanEngelen 

I had to fix several issues for this PR to succeed
- There was an issue with the pooch cache. The teamcity admin created a symlink which is giving permission errors. I fixed it by settings the IMOD_DATA_DIR environment variable and moving the cache into the buildfolder
- There was a data race between multiple pytest nodes accessing the same file. This has been resolved by adding a filelock
- Several MyPy errors have been resolved
- pip has been added as a dependency

JoerivanEngelen (approved):
Looks good, thanks for the fixes! The file lock seems like an improvement to me :-)

PR #1165, opened 2024-08-21
Issue #1107 NumPy 2.0 migration

### LINKED ISSUE #1107: Migrate to numpy 2.x
Numpy 2.x has been released, with quite some breaking changes. The migration guide provides some useful information:

https://numpy.org/devdocs/numpy_2_0_migration_guide.html#numpy-2-migration-guide

We have to investigate if we can be compatible soon.

Manangka:
Fixes #1107

# Description
This commit contains changes required to migrate to NumPy 2.0

# Checklist
- [ ] Links to correct issue
- [ ] Unit tests were added


imod/prepare/spatial.py
```diff
@@ -292,6 +294,8 @@ def _handle_dtype(dtype, nodata):
         "int32": (-2147483648, 2147483647),
         "float32": (-3.4028235e38, 3.4028235e38),
         "float64": (-1.7976931348623157e308, 1.7976931348623157e308),
+        "uint64": (0, 18446744073709551615),
+        "int64": (-9223372036854775808, 9223372036854775807),
```
JoerivanEngelen, line 298: Nice that this got an update, I infer from rasterio source code that this has been introduced in GDAL 3.5; GDAL is quite notorious for contributing to dependency hell in Python projects in the past, though it has recently gotten a lot better. 

I think to be on the safe side, we should either add a GDAL version check here (like rasterio does) or restrict ourselves to GDAL 3.5 or higher in the pixi or pyproject.toml. 

Versionwise, what is common now:
GDAL 3.5 was released in may 2022. I see rasterio requires GDAL 3.3 or higher in their setup.py. So I think it is fine to enforce GDAL 3.5 or higher.

JoerivanEngelen (approved):
Looks good, and thanks for cleaning up some outdated parts in prepare.spatial. I have one comment on the minimal GDAL versions now required. Could you add those to the pyproject.toml and pixi.toml?

imod/prepare/spatial.py
Manangka: Restricted the version of GDAL to a minimum of 3.5

JoerivanEngelen:
@Manangka I see one of the examples failed, can you investigate?

PR #1186, opened 2024-09-04
Issue #1176 assign wells with point filters

### LINKED ISSUE #1176: Assign wells: Wells with a filter length of zero
I haven't tested this yet, but I think our ``assign_wells`` function is not able to allocate wells with zero filter length. 
In this case iMOD5 directly allocates these wells to the layer, I have to verify if our function does that as well, but I couldn't find any specific logic for it in my quick glance.

Validation at least throws an error now, which it shouldn't. 

Furthermore, ideally, I'd would like to align ``assign_wells`` with ``imod/prepare/topsystem/conductance.py``. To support the 5 methods that iMOD5 to redistribute well rates.

JoerivanEngelen:
Fixes #1176 

# Description

- Assign wells with point filters (filters with a zero length) to layers. Before, these would be removed in the compute_overlap method.
- Add test case for wells with a point filter
- Set validation so that screen_top can equal screen_bottom

# Checklist

- [x] Links to correct issue
- [x] Unit tests were added

### DIFF
--- a/imod/prepare/wells.py
+++ b/imod/prepare/wells.py
@@ -31,8 +31,11 @@ def compute_point_filter_overlap(bounds_wells, bounds_layers):
     thickness.
     """
     zero_filter_length = bounds_wells[:, 1] == bounds_wells[:, 0]
+    in_layer = (bounds_layers[:, 1] > bounds_wells[:, 1]) & (
+        bounds_layers[:, 0] < bounds_wells[:, 0]
+    )
     layer_thickness = bounds_layers[:, 1] - bounds_layers[:, 0]
-    return zero_filter_length.astype(float) * layer_thickness
+    return zero_filter_length.astype(float) * in_layer.astype(float) * layer_thickness
 
 
 def compute_overlap(wells, top, bottom):
@@ -58,6 +61,7 @@ def compute_overlap(wells, top, bottom):
         layer_bounds,
     )
     return np.maximum(interval_filter_overlap, point_filter_overlap)
+
 
 def locate_wells(
     wells: pd.DataFrame,
--- a/imod/prepare/wells.py
+++ b/imod/prepare/wells.py
@@ -175,7 +175,7 @@ def assign_wells(
 
     # Distribute rate according to transmissivity.
     n_layer, n_well = xy_top.shape
-    df = pd.DataFrame(
+    df_factor = pd.DataFrame(
         index=pd.Index(np.tile(first.index, n_layer), name="id"),
         data={
             "layer": np.repeat(top["layer"], n_well),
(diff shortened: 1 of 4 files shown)


imod/prepare/wells.py
```diff
@@ -25,21 +25,42 @@ def vectorized_overlap(bounds_a, bounds_b):
     )
 
 
+def compute_point_filter_overlap(bounds_wells, bounds_layers):
+    """
+    Special case for filters with zero filter length, these are set to layer
+    thickness.
+    """
+    zero_filter_length = bounds_wells[:, 1] == bounds_wells[:, 0]
```
Manangka, line 33: Please start a boolean variable with "is" or "has" ->` is_zero_length_filter` or `has_zero_filter_length`

imod/prepare/wells.py
```diff
@@ -25,21 +25,42 @@ def vectorized_overlap(bounds_a, bounds_b):
     )
 
 
+def compute_point_filter_overlap(bounds_wells, bounds_layers):
```
Manangka, line 28: Could you add type hints?

It makes it a bit hard to review not knowing what the input and output of this method is

imod/prepare/wells.py
```diff
@@ -25,21 +25,42 @@ def vectorized_overlap(bounds_a, bounds_b):
...
+    """
+    Special case for filters with zero filter length, these are set to layer
+    thickness.
+    """
+    zero_filter_length = bounds_wells[:, 1] == bounds_wells[:, 0]
+    in_layer = (bounds_layers[:, 1] >= bounds_wells[:, 1]) & (
```
Manangka, line 34: Could you unwrap the bounds_layer and bounds_wells arrays? Im guessing one of them corresponds to the tops and the other to the bottoms

imod/prepare/wells.py
```diff
@@ -25,21 +25,42 @@ def vectorized_overlap(bounds_a, bounds_b):
...
+        bounds_layers[:, 0] < bounds_wells[:, 0]
+    )
+    layer_thickness = bounds_layers[:, 1] - bounds_layers[:, 0]
+    return zero_filter_length.astype(float) * in_layer.astype(float) * layer_thickness
```
Manangka, line 38: Could the following snippet also work?
`zero_filter_lengths = layer_thickness[zero_filter_length & in_layer]`

Manangka (changes requested):

imod/prepare/wells.py
JoerivanEngelen: Improved to ``has_zero_filter_length``.

JoerivanEngelen: Improved to ``has_zero_filter_length``.

JoerivanEngelen:
The request for type hints lead to some extra required type annotations to be implemented for MyPy. A no-regret, but the slight downside that this changeset now influences more places of the code.

Manangka (approved):

imod/prepare/wells.py
Manangka: Thanks! This makes it way easier for me to follow the logic

PR #1189, opened 2024-09-05
Issue #1188 Update the numba_celltree package

### LINKED ISSUE #1188: Update numba_celltree package
In #1165 the numba_celltree package has been pinnend to a version lower then 0.2.
The reason for this is that a change has been made in that package which affect the locate_points method.
https://github.com/Deltares/numba_celltree/blame/4342ac78481a7e298f5cc05aa211f0bf5532e5ab/numba_celltree/query.py#L28

As a result the following tests fail:
- test_in_bounds
- test_get_indices__unstructured,
- test_get_values__unstructured

@Huite suggested to alter [`in_bounds &= (x >= xmin) & (x < xmax)`](https://github.com/Deltares/imod-python/blob/94029916b5ad11364f9369e4db10c6a7c013c463/imod/select/points.py#L115) into `in_bounds &= (x >= xmin) & (x <= xmax)`

As a consequence other tests will fail and need to be fixed

After this issue the line `numba_celltree = "<0.2"` can be removed from the pixi.toml file

Manangka:
Fixes #1188

# Description
In the numba_celltree package a change has been made that changes the behavior of the locate_points method.

In structured grids the behavior is that when a point lies on the left or bottom edge it is considered to be in the cell.
When the point lies on the right or top edge its considered outside of the cell

This kind of logic doesn't work for unstructured grids because the orientation of the cell can be anything.
However there are unit tests that do assume that this applies for unstructured grids as well. And those are the tests that are failing right now.

To solve this the bahavior has changed such that when a point lies on a cell edge its is considered to be in the cell. This also applies to structred grids. 

This was

# Checklist

- [ ] Links to correct issue
- [ ] Unit tests were added

### DIFF
--- a/imod/select/points.py
+++ b/imod/select/points.py
@@ -112,7 +112,7 @@ def points_in_bounds(da, **points):
         da_x = da.coords[key]
         _, xmin, xmax = imod.util.spatial.coord_reference(da_x)
         # Inplace bitwise operator
-        in_bounds &= (x >= xmin) & (x < xmax)
+        in_bounds &= (x >= xmin) & (x <= xmax)
 
     return in_bounds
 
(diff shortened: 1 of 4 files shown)


JoerivanEngelen (approved):
Thanks for fixing this! This made me realize we have an edge case for well placement on a partition border, but that was already a problem for unstructured grids.

PR #1211, opened 2024-09-12
Issue #1209 bypass value error well assign

### LINKED ISSUE #1209: Add option to bypass `ValueError` for removed wells during well assigning in ``to_mf6_pkg``
``to_mf6_pkg`` now throws a ValueError when wells are removed entirely during their assigning to model cells, which cannot be bypassed. Even worse: The error is thrown before erronous wells' locations are logged.

- There currently is an ``is_partitioned`` flag which turns off the check, but that can be perhaps renamed to something to be used outside the partitioning context. 
- Logging should be done before error is thrown
- Error message can include to ids of removed wells and point to option to turn off the check

JoerivanEngelen:
Fixes #1209

# Description
- Add setting to ``to_mf6_pkg``: ``error_on_well_removal``, which throws an error if wells are removed entirely during well assignment.
- Separated code to collect filtered well ids in separate method. This method is now also called to collect ids of errors being removed in error message.
- Added attribute ``_is_from_imod5`` to Modflow6Simulation to keep track if the simulation is loaded from iMOD5 data or not. Based on this, some validation settings might be set to be a bit more permissive. (Or call some extra cleanup?)
- ``is_from_imod5`` property to WriteContext to keep track of the same thing within the writing context.

# Checklist
- [x] Links to correct issue
- [x] Unit tests were added


imod/mf6/simulation.py
```diff
@@ -97,6 +97,7 @@ def __init__(self, name):
         self.name = name
         self.directory = None
         self._initialize_template()
+        self._is_from_imod5 = False
```
Manangka, line 100: Hmm imod5 converter logic is now being mixed up with the main logic. I'm not sure this is a good thing

imod/mf6/wel.py
```diff
@@ -429,19 +434,21 @@ def to_mf6_pkg(
...
-            raise ValueError("No wells were assigned in package. None were present.")
+            raise ValidationError(
+                "No wells were assigned in package. None were present."
+            )
 
-        if not is_partitioned and nwells_df != nwells_assigned:
-            raise ValueError(
-                "One or more well(s) are completely invalid due to minimum conductivity and thickness constraints."
+        wells_assigned = self._assign_wells_to_layers(wells_df, active, top, bottom, k)
+        filtered_ids_in_assign = self.gather_filtered_well_ids(wells_assigned, wells_df)
```
Manangka, line 443: I would rename the above variables to:
`assigned_wells`
`filtered_assigned_wells` or `filtered_wells_ids`
I think that reads a bit easier

imod/mf6/wel.py
```diff
@@ -429,19 +434,21 @@ def to_mf6_pkg(
...
+        message_assign = self.to_mf6_package_information(
+            filtered_ids_in_assign, reason_text="permeability/thickness constraints"
+        )
+        if error_on_well_removal and len(filtered_ids_in_assign) > 0:
+            logger.log(
+                loglevel=LogLevel.ERROR, message=message_assign, additional_depth=2
```
Manangka, line 449: Why did you add `additional_depth=2`? I don't think it is necessary here. 

The `additional_depth` parameter is used when you want to log the file and line level of one level lower then the current line. This comes in handy for the logging decorator because you don't want to the log the line in the decrator but the line in the method that is decorated

Manangka (changes requested):

imod/mf6/simulation.py
JoerivanEngelen: Will be picked up further in https://github.com/Deltares/imod-python/issues/1222

imod/mf6/wel.py
JoerivanEngelen: Woops, relict of some copy pasta action (I didn't know what the ``additional_depth`` parameter did.) 

Fixed!

imod/mf6/wel.py
JoerivanEngelen: Done

imod/mf6/wel.py
```diff
@@ -439,37 +439,35 @@ def to_mf6_pkg(
...
 
-        wells_assigned = self._assign_wells_to_layers(wells_df, active, top, bottom, k)
-        filtered_ids_in_assign = self.gather_filtered_well_ids(wells_assigned, wells_df)
+        assigned_wells = self._assign_wells_to_layers(wells_df, active, top, bottom, k)
+        filterd_assigned_well_ids = self.gather_filtered_well_ids(
+            assigned_wells, wells_df
+        )
         message_assign = self.to_mf6_package_information(
-            filtered_ids_in_assign, reason_text="permeability/thickness constraints"
+            filterd_assigned_well_ids, reason_text="permeability/thickness constraints"
```
Manangka, line 447: Typo "filtered"

Manangka (approved):

PR #1231, opened 2024-09-18
Issue #1230 distribut drn elevation equals surface level

### LINKED ISSUE #1230: distribute conductances edge case: when surface level equals drain elevation, conductance set to nan
I encountered an edge case with the LHM, when drain elevation equals surface level, the conductance is set to nan.
Also reproduced this with a small unittest

Probably happens in ``_compute_crosscut_thickness``

JoerivanEngelen:
Fixes #1230

# Description
- Fix bug where distributed conductances were dropped when their elevation equalled surface level
- In topsystem test cases, make sure "layer" coordinate is dropped consistently amongst cases.

# Checklist

- [x] Links to correct issue
- [x] Unit tests were added


JoerivanEngelen:
The unittest seem to fail on something different, it appears that the FileLocks don't fully work?

PR #1238, opened 2024-10-04
Use apply_ufunc in `imod.prepare.fill`. Fixes #1237

### LINKED ISSUE #1237: `imod.prepare.fill` assign to the result of a `sel`
This is fully a sin of my previous self, but:

```python
    out = xr.full_like(da, np.nan)
    if invalid is None:
        invalid = np.isnan(da)
    if by:
        for coordvalue in da[by]:
            d = {by: coordvalue}
            out.sel(d)[...] = _fill_np(da.sel(d).values, invalid.sel(d).values)
    else:
        out.values = _fill_np(da.values, invalid.values)

    return out
```

This isn't good. On another note, `xarray.apply_ufunc` is a much better solution for for the logic here.

Huite:
Fixes #1237.

This is unfortunately a breaking change, but keeping the previous API makes the logic a lot more complicated (and arguably makes the function worse).

For example, keeping the `invalid` (might need a better name too) as a DataArray almost surely requires including it in the broadcasting logic of `apply_ufunc`.

### DIFF
--- a/docs/api/changelog.rst
+++ b/docs/api/changelog.rst
@@ -6,6 +6,23 @@ All notable changes to this project will be documented in this file.
 The format is based on `Keep a Changelog`_, and this project adheres to
 `Semantic Versioning`_.
 
+Unreleased
+----------
+
+Fixed
+~~~~~
+
+- :func:`imod.prepare.fill` previously assigned to the result of an xarray
+  ``.sel`` operation. This might not work for dask backed data and has been
+  fixed.
+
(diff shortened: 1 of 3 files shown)


JoerivanEngelen (commented):
Thanks for the fix, I'm always fond of outsourcing logic :-) , could you also update the changelog (since we are changing the public API)?

JoerivanEngelen (approved):

PR #1239, opened 2024-10-04
Use apply_ufunc. Remove pcg module, use scipy cg instead.

Huite:
This simplifies considerably, and makes the interpolation more flexible as well via the use of `apply_ufunc`.

Unfortunately, it is a breaking change, but a number of arguments don't make sense (or never did).

This also introduces a dependency on the preconditioner from xugrid.


JoerivanEngelen (approved):
I've finished the PR, had to update a unittest and an example, and extended the changelog.

Issue #159, opened 2021-03-07
mf6.open_cbc does not return the right lower face face in a 2D(layer, x) model?


Manangka:
In GitLab by @Huite on Mar 7, 2021, 21:34

Checking the result of a 2D example, I'm getting 0 values in the lower face flow, while I have only one row defined.
I didn't test the `_dis_indices` function for 2D arrays, it might not be returning the right numbers here ...

Manangka:
In GitLab by @Huite on Jun 17, 2021, 09:19

Solution requires passing the shape of the arrays (nlay, nrow, ncol) from the grb file to this function.

JoerivanEngelen:
Fixed in #1243

PR #1243, opened 2024-10-17
Issue #1242 Avoid swapping of FRF and FFF in open_cbc for 2D grids

### LINKED ISSUE #1242: mf6.open_cbc: FRF and FFF are swapped in case a model has only one column
In case a model has one column and multiple rows, the flux-ja-face conversion to FRF, FFF goes wrong in the dis_indices method. 

For this case the + 1 node does not correspond to the right neighbor but the front one. Since the + 1 and + ncol evaluation are both true, the order of evaluation causes the swap to FRF. 

Should be solved by evaluating d == 1 & ncol != 1

HendrikKok:
Fixes #1242 and #159

Fixes swapping budget output:

Case 1: FRF to FFF in cases ncol == 1
Case 2: FLF to FFF  in case numer of cell per layer == 1

Added tests for single column model (2D) and single column and row (1D) model.

# Description

# Checklist

- [x] Links to correct issue
- [x] Unit tests were added

### DIFF
--- a/docs/api/changelog.rst
+++ b/docs/api/changelog.rst
@@ -12,8 +12,9 @@ Unreleased
 Fixed
 ~~~~~
 
-- Fix `imod.mf6.open_cbc`: it failed with ``flowja=False`` on budget output for
+- `imod.mf6.open_cbc` failing with ``flowja=False`` on budget output for
   DISV models if the model contained inactive cells.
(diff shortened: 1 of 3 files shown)


HendrikKok:
@JoerivanEngelen. Last commit should fix issue #159 as well.

PR #1256, opened 2024-10-25
Issue #728 metaswap grid agnostic well

### LINKED ISSUE #728: Modflow-MetaSWAP coupler mappings should use grid-agnostic wells or well adapter
The ``msw.CouplerMapping`` , ``msw.Sprinkling``, and ``couplers.MetaMod`` (in ``primod``) currently use the deprecated ``mf6.WellDisStructured`` object. We should start supporting the new architecture. 

This will also require updates to:

- [primod](https://github.com/Deltares/imod_coupler/tree/main/pre-processing) because coupler logic is now placed there
- The example in the [iMOD Documentation](https://github.com/Deltares/iMOD-Documentation) 

We might be able to use the ``Mf6Well`` adapter directly here:
+ Less work to update existing ``_render`` method
+ No mf6 data needs to be provided to render method. With grid agnostic packages, this would be  (``idomain``, ``bottom``, ``top``, ``k``)
- However, this makes it harder to clip and regrid this package object this would require the ``regrid_like`` to update the indices

The alternative is providing the grid agnostic ``Well`` package: 
+ Easy to clip & regrid
- This requires extra information to the  ``_render`` method, all contained in the ``GroundwaterFlowModel`` object. 

The third option is creating a GridWell package: A grid of well rates, comparable to other boundary conditions.
+ Behaves similair to other packages
- Quite a change in public API: Would users like this?

Also: Since MODFLOW6 models have an idomain correction after regridding (some NPF cells may become inactive at edges, where idomain cells were not inactive), we always have to regrid the GroundwaterFlowModel first, before regridding MetaSWAP model.

JoerivanEngelen:
Fixes #728

# Description
This PR updates two MetaSWAP packages that depend on wells to use of the deprecated ``WellDisStructured`` object to the ``Mf6Wel`` object.

It fixes the first small step of the big metaswap refactor. I chose a feature branch as base branch to merge into, to still have the option to create some hotfixes for a new release of iMOD Python while primod is being updated.

I also renamed some variable names and dimension names with clearer names. Furthermore, I type annotated some method.

# Checklist

- [x] Links to correct issue
- [ ] Unit tests were added

### DIFF
--- a/imod/mf6/wel.py
+++ b/imod/mf6/wel.py
@@ -218,6 +218,78 @@ def get_all_imod5_prj_well_times(imod5_data: dict) -> list[datetime]:
     return sorted(set(wel_times_flat))
 
 
+def derive_cellid_from_points(
+    dst_grid: GridDataArray,
+    x: list,
+    y: list,
+    layer: list,
+) -> GridDataArray:
+    """
+    Create DataArray with Modflow6 cell identifiers based on x, y coordinates
+    in a dataframe. For structured grid this DataArray contains 3 columns:
+    ``layer, row, column``. For unstructured grids, this contains 2 columns:
+    ``layer, cell2d``.
+    See also: https://water.usgs.gov/water-resources/software/MODFLOW-6/mf6io_6.4.0.pdf#page=35
+
+    Note
+    ----
+    The "layer" coordinate should already be provided in the dataframe.
+    To determine the layer coordinate based on screen depts, look at
+    :func:`imod.prepare.wells.assign_wells`.
+
+    Parameters
+    ----------
+    dst_grid: {xr.DataArray, xu.UgridDataArray}
+        Destination grid to map the points to based on their x and y coordinates.
+    x: {list, np.array}
+        array-like with x-coordinates
+    y: {list, np.array}
+        array-like with y-coordinates
+    layer: {list, np.array}
+        array-like with layer-coordinates
+
+    Returns
+    -------
+    cellid : xr.DataArray
+        2D DataArray with a ``ncellid`` rows and 3 to 2 columns, depending
+        on whether on a structured or unstructured grid."""
+
+    # Find indices belonging to x, y coordinates
+    indices_cell2d = points_indices(dst_grid, out_of_bounds="ignore", x=x, y=y)
+    # Convert cell2d indices from 0-based to 1-based.
+    indices_cell2d = {dim: index + 1 for dim, index in indices_cell2d.items()}
+    # Prepare layer indices, for later concatenation
+
+    if isinstance(dst_grid, xu.UgridDataArray):
+        indices_layer = xr.DataArray(
+            layer, coords=indices_cell2d["mesh2d_nFaces"].coords
+        )
+        face_dim = dst_grid.ugrid.grid.face_dimension
+        indices_cell2d_dims = [face_dim]
+        cell2d_coords = ["cell2d"]
+    else:
+        indices_layer = xr.DataArray(layer, coords=indices_cell2d["x"].coords)
+        indices_cell2d_dims = ["y", "x"]
+        cell2d_coords = ["row", "column"]
+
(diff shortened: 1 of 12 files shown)


HendrikKok (approved):

imod/msw/sprinkling.py
```diff
@@ -55,23 +55,29 @@ def __init__(
...
+            raise TypeError(rf"well not of type 'Mf6Wel', got '{type(well)}'")
         self.well = well
 
         self._pkgcheck()
 
     def _render(self, file, index, svat):
-        well_row = self.well["row"] - 1
-        well_column = self.well["column"] - 1
-        well_layer = self.well["layer"]
+        cellid = self.well["cellid"]
```
Manangka, line 70: You could rename this to `well_cellid`

imod/msw/sprinkling.py
```diff
@@ -55,23 +55,29 @@ def __init__(
...
+        if len(cellid.coords["nmax_cellid"]) != 3:
+            raise TypeError("Coupling to unstructured grids is not supported.")
+
+        well_layer = cellid.sel(nmax_cellid="layer").data
```
Manangka, line 74: What is `nmac_cellid`? I cant figure it our by just looking at this file
Could you rename it to something more describing?

imod/mf6/wel.py
```diff
@@ -248,7 +320,7 @@ def _create_cellid(
         df_for_cellid = assigned_wells.groupby("index").first()
```
Manangka, line 320: Would unique_assigned_wells be a better name?

imod/mf6/wel.py
```diff
@@ -248,7 +320,7 @@ def _create_cellid(
...
         d_for_cellid = df_for_cellid[["x", "y", "layer"]].to_dict("list")
 
-        return self._derive_cellid_from_points(like, **d_for_cellid)
+        return derive_cellid_from_points(like, **d_for_cellid)
```
Manangka, line 323: If the well is ever moved the cellid needs to be updated. 
Thus method then needs to be called whenever the row,column or layer is adjusted

I don't know how likely that is but i cant imagine a use cage where you start with a base model with wells and then move some of them to see the effect of it.

imod/mf6/wel.py
```diff
@@ -248,7 +320,7 @@ def _create_cellid(
         df_for_cellid = assigned_wells.groupby("index").first()
         d_for_cellid = df_for_cellid[["x", "y", "layer"]].to_dict("list")
```
Manangka, line 321: I find the name cellid confusing. In MODFLOW terms this is the row, column, layer or the 2dcellid, layer.
In this case it would call it something like `id`, `identifier`, `name` or `well_id`

Manangka (changes requested):

imod/msw/sprinkling.py
JoerivanEngelen: I based this name on an example I found in the UGRID conventions. There it is named ``nMaxMesh2_face_nodes``, meaning max amount of nodes per face. The "max" is relevant in Ugrids, as meshes can be flexible so the amount of nodes per face can vary in the grid. 

In our case this is not happening, so the "max" is wrong here. I'll rename to ``ncellid``, as that is a better name for this dimension.

imod/mf6/wel.py
JoerivanEngelen: But row, column, layer is what I refer to with the ``cellid``. At some point in the future we might also have to support 2dcellid, but that will probably not happen soon.

imod/msw/sprinkling.py
Huite: +1 for `ndim_cellid`

JoerivanEngelen: After a brief discussion with @Manangka, I renamed it to ``dim_cellid``.

Manangka (approved):

Issue #1268, opened 2024-11-01
Bug: Modflow6Model.__get_k cannot be called when NPF package is assigned to model with another key than "npf"


JoerivanEngelen:
Modflow6Model.__get_k cannot be called when NPF package is assigned to model with another key than "npf"

Calls to __get_k when allocating Well, LayeredWell, and HorizontalFlowBarrier objects, so this will be problematic then.
Instead, ``_get_pkgid`` has to be called, like is done for example in ``__get_domain_geometry`` as well.

PR #1269, opened 2024-11-01
Fix Issue #1267 and #1268 prepare mf6 well model

### LINKED ISSUE #1267: Add method to GroundwaterFlowModel to convert GridAgnosticWell to Mf6Wel
https://github.com/Deltares/imod-python/issues/1255 introduced that the ``MetaSwapModel`` depended on ``Mf6Wel`` objects.

However, generating these objects is tedious, as data from the dis, and npf package are required. Therefore, adding a method to prepare wells from the Modflow6Model object helps. This method can also be called in ``primod`` to fetch Mf6Wel objects.

JoerivanEngelen:
Fixes #1267 and #1268

# Description
This does the following:
- fixes a bug with ``self.__get_k`` and different keys for the ``NodePropertyFlow``. Now the method ``__get_k`` is implicitly hardcoded with the "npf" key, as this is the ``NodePropertyFlow._pkg_id``
- Adds a method ``Groundwater.prepare_wel_for_mf6`` to allocate wells to the public API. Users can call this for debugging purposes and ``primod`` can call it to fetch Mf6Wel objects to for its coupling schemes
- Make ``_get_k`` and ``_get_domain_geometry`` methods semi-private and add tests for these functions.

# Checklist

- [x] Links to correct issue
- [x] Unit tests were added

### DIFF
--- a/docs/api/changelog.rst
+++ b/docs/api/changelog.rst
@@ -27,6 +27,15 @@ Added
 ~~~~~
 
 - :meth:`imod.msw.MetaSwapModel.regrid_like` to regrid MetaSWAP models.
+- :meth:`imod.mf6.GroundwaterFlowModel.prepare_wel_for_mf6` to prepare wells for
+  MODFLOW6, for debugging purposes.
+
+Fixed
+~~~~~
+
+- Bug where error was thrown when :class:`imod.mf6.NodePropertyFlow` was
+  assigned to :class:`imod.mf6.GroundwaterFlowModel` with key different from
+  ``"npf"`` upon writing, as well as well or horizontal flow barrier packages.
 
 
 [Unreleased]
--- a/docs/api/changelog.rst
+++ b/docs/api/changelog.rst
@@ -35,7 +35,7 @@ Fixed
 
 - Bug where error was thrown when :class:`imod.mf6.NodePropertyFlow` was
   assigned to :class:`imod.mf6.GroundwaterFlowModel` with key different from
-  ``"npf"`` upon writing, as well as well or horizontal flow barrier packages.
+  ``"npf"`` upon writing, along with well or horizontal flow barrier packages.
 
 
 [Unreleased]
--- a/imod/mf6/model.py
+++ b/imod/mf6/model.py
@@ -243,9 +243,6 @@ class Modflow6Model(collections.UserDict, IModel, abc.ABC):
     def prepare_wel_for_mf6(
         self, pkgname: str, validate: bool = True, strict_well_validation: bool = True
     ) -> Mf6Wel:
-        validate_context = ValidationContext(
-            validate=validate, strict_well_validation=strict_well_validation
-        )
         """
         Prepare grid-agnostic well for MODFLOW6, using the models grid
         information and hydraulic conductivities. Allocates LayeredWell & Well
@@ -272,6 +269,9 @@ class Modflow6Model(collections.UserDict, IModel, abc.ABC):
             Direct representation of MODFLOW6 WEL package, with 'cellid'
             indicating layer, row columns.
         """
+        validate_context = ValidationContext(
+            validate=validate, strict_well_validation=strict_well_validation
+        )
         return self._prepare_wel_for_mf6(pkgname, validate_context)
 
     @standard_log_decorator()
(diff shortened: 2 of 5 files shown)


docs/api/changelog.rst
```diff
@@ -27,6 +27,15 @@ Added
...
 - :meth:`imod.msw.MetaSwapModel.regrid_like` to regrid MetaSWAP models.
+- :meth:`imod.mf6.GroundwaterFlowModel.prepare_wel_for_mf6` to prepare wells for
+  MODFLOW6, for debugging purposes.
+
+Fixed
+~~~~~
+
+- Bug where error was thrown when :class:`imod.mf6.NodePropertyFlow` was
+  assigned to :class:`imod.mf6.GroundwaterFlowModel` with key different from
+  ``"npf"`` upon writing, as well as well or horizontal flow barrier packages.
```
HendrikKok, line 38: typo

imod/mf6/model.py
```diff
@@ -239,6 +240,60 @@ def validate(self, model_name: str = "") -> StatusInfoBase:
 
         return model_status_info
 
+    def prepare_wel_for_mf6(
+        self, pkgname: str, validate: bool = True, strict_well_validation: bool = True
+    ) -> Mf6Wel:
+        validate_context = ValidationContext(
+            validate=validate, strict_well_validation=strict_well_validation
+        )
+        """
```
HendrikKok, line 249: Did you move the docstring down intentionally?

imod/tests/test_mf6/test_mf6_model.py
```diff
@@ -370,3 +371,55 @@ def test_deepcopy(
...
+    assert isinstance(k, xr.DataArray)
+
+
+def test_get_domain_geometry(structured_flow_model: GroundwaterFlowModel):
+    # Act
+    top, bottom, idomain = structured_flow_model._get_domain_geometry()
+    # Assert
+    assert isinstance(top, xr.DataArray)
+    assert isinstance(bottom, xr.DataArray)
+    assert isinstance(idomain, xr.DataArray)
```
HendrikKok, line 425: There is no need to do a check on the values?

HendrikKok (commented):
Nice work. I have some minor comments

imod/mf6/model.py
```diff
@@ -190,14 +191,14 @@ def __get_domain_geometry(
...
-        except RuntimeError:
-            raise ValidationError("expected one package of type ModePropertyFlow")
-
-        k = npf["k"]
-        return k
+    def _get_k(self):
+        npf_key = self._get_pkgkey(imod.mf6.NodePropertyFlow._pkg_id)
+        if not npf_key:
+            raise KeyError(
+                "No NodePropertyFlow package was assigned to Modflow6Model, make sure this is included."
```
Manangka, line 198: Would the following error message be better?
`Unable to obtain the hydraulic conductivity. Make sure a NodePropertyFlow package isassigned to the Modflow6Model`

imod/mf6/model.py
```diff
@@ -274,15 +329,7 @@ def write(
...
-                    top, bottom, idomain = self.__get_domain_geometry()
-                    k = self.__get_k()
-                    mf6_well_pkg = pkg._to_mf6_pkg(
-                        idomain,
-                        top,
-                        bottom,
-                        k,
-                        validate_context,
-                    )
+                    mf6_well_pkg = self._prepare_wel_for_mf6(pkg_name, validate_context)
```
Manangka, line 332: The _prepare_wel_for_mf6 isnt called from anywhere else is it?
There is now a double check for the pkgtype. On line 331 and on line 282, I would get rid of one of them.

In fact i don't think that lines 283 to 286 will ever be reached

JoerivanEngelen: Context: It is useful for users to have a public api method with which they can ask an Mf6Wel object from the modflow model. This is ``prepare_wel_for_mf6``. Internally, we use the ValidationContext object, which we pass on to other methods. However, generating these context object (also WriteContext), is a bit obscure to users. Instead, I created a public API function, which generates this for users, and then passes this context on.

As the method can now be called outside from the `write``, I had to add an extra type check. It is unnecessary when writing a model, but it can be reached when passing on the wrong string to ``prepare_wel_for_mf6``.

imod/mf6/model.py
JoerivanEngelen: Yes that is better! Implemented that message.

PR #1274, opened 2024-11-06
Issue #1273 cleanup example

### LINKED ISSUE #1273: Add example showcasing cleanup utilities
Cleanup utilities were added for #1164.

Unfortunately, I didn't find time yet to add an example for this new functionality. It is good to have some examples, also for the Delft Software Days.

JoerivanEngelen:
Fixes #1273

# Description
Add example of cleanup utilities to the user guide. I think it shows a common use-case for cleaning up data.

Technically the most challenging issue was to work around the fact that sphinx_gallery doesn't have an option to select expected errors in specific code blocks. There is [this PR](https://github.com/sphinx-gallery/sphinx-gallery/pull/930) to introduce this, but unfortunately for us this has not been finished yet. 

I resorted to trapping the error and printing it with a context manager, that seemed like the cleanest solution to me. I added this small context to the public API, so that users can also read in the online documentation what the context manager does.

# Checklist

- [x] Links to correct issue
- [x] Unit tests were added

### DIFF
--- a/examples/user-guide/10-cleanup.py
+++ b/examples/user-guide/10-cleanup.py
@@ -19,21 +19,33 @@ gwf_simulation = imod.data.hondsrug_simulation(tmpdir / "hondsrug_saved")
 # %%
 
 def get_colleagues_data(gwf_model):
-    from copy import deepcopy
-    riv_ds = deepcopy(gwf_model["riv"].dataset)
+    import xarray as xr
+    dis_ds = gwf_model["dis"].dataset
+    riv_ds_old = gwf_model["riv"].dataset
+    # Existing RIV package has only layer coord with [1, 3, 5, 6]. This causes
+    # problems with some methods, at least with cleanup. Therefore align data
+    # here for this example. Not sure if we want to support river packages with
+    # limited layer coords.
+    riv_ds, _ = xr.align(riv_ds_old, dis_ds, join="outer")
     x = riv_ds.coords["x"]
+    y = riv_ds.coords["y"]
     riv_bot_da = riv_ds["bottom_elevation"]
     riv_ds["stage"] += 0.05
     riv_ds["stage"] = riv_ds["stage"].where(x > 239500)
     riv_ds["conductance"] = riv_ds["conductance"].fillna(0.0)
+    x_preserve = (x < 244200) | (x > 246000)
+    y_preserve = (y < 560000) | (y > 561000)
     riv_ds["bottom_elevation"] = riv_bot_da.where(
-        (x > 244000) & (x < 246000), 
+        x_preserve | y_preserve, 
     riv_bot_da + 0.15)
     return riv_ds
 
 
 
 # %%
+#
+# Update existing model with new data
+# -----------------------------------
 #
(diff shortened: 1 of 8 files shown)


examples/user-guide/10-cleanup.py
```diff
@@ -0,0 +1,245 @@
+"""
+Data cleanup
+============
+
+More often than not, data contained in databases is not entirely consistent,
+causing errors. It therefore is useful to have some utilities at hand to clean
+up data. We included some convenience methods to help cleaning up inconsistent
```
Manangka, line 7: cleaning up => clean up

examples/user-guide/10-cleanup.py
```diff
@@ -0,0 +1,245 @@
...
+
+gwf_simulation = imod.data.hondsrug_simulation(tmpdir / "hondsrug_saved")
+
+
+# %%
+#
+# Update existing model with new data
+# -----------------------------------
+#
+# Your dear colleague has brought you some new data for the river package, which
```
Manangka, line 36: I like this :D

examples/user-guide/10-cleanup.py
```diff
@@ -0,0 +1,245 @@
...
+#
+# iMOD Python has a set of utilities to help you with cleaning erronous data.
+# Note that these cleanup utilities can only fix a limited set of common
+# mistakes, in ways you might not need/like. It therefore is always wise to
+# verify if you think the fixes are correct in this case.
+#
+# Let's clean up the problematic river data using the
+# :meth:`imod.mf6.River.cleanup` method. Note that this method changes the data
+# inplace. This means that the package's will be updated, without returning a
+# new copy of the package. This means that do comparisons of the
```
Manangka, line 144: "that do" = > "that to do"

imod/data/sample_data.py
```diff
@@ -248,3 +248,30 @@ def hondsrug_planar_river() -> xr.Dataset:
...
+    """
+    River data with some mistakes introduced to showcase cleanup
+    utilities.
+    """
+    gwf_model = hondsrug_simulation(path)["GWF"]
+    dis_ds = gwf_model["dis"].dataset
+    riv_ds_old = gwf_model["riv"].dataset
+    # Existing RIV package has only layer coord with [1, 3, 5, 6]. This causes
+    # problems with some methods, at least with cleanup. Therefore align data
+    # here for the cleanup example. Not sure if we want to support river
```
Manangka, line 263: "Not sure if we want to support river packages with limited layer coords"

I'm not sure if we should put the above in the comment.

Instead you could also write:
"River packages with limited layer coords are currently not supported."

Manangka:
I really like this way of writing examples!!

imod/data/sample_data.py
JoerivanEngelen: I agree, I changed it to "River packages with limited layer coords are currently not fully supported.", as you can apparently write a functional model with the river package in only limited layers, but cannot call cleanup on it. I wasn't sure if we want to support that to begin with.

PR #1278, opened 2024-11-08
Issue #1223 Remove WriteContext from public api

### LINKED ISSUE #1223: Remove ``WriteContext`` from public API
Somewhere around https://github.com/Deltares/imod-python/commit/5523c2425fe28bf618c20d9b7de624b4e74ccc6f the WriteContext was introduced, this is an object to collect settings for writing (directory, save as binary, etc.). This is generated internally when calling ``Modflow6Simulation.write``. The ``WriteContext`` is passed through from Simulation to Model to Package, and accepted as argument for the ``Package.write`` class. However, this altered the public API, as now users have to first create a ``WriteContext``, an object stowed away in a non-public module, to be able to write their packages separately. 

I'm not sure how large the use-case is, but it the write context is quite an extra hurdle for users now to call a method that is part of the public API. I like the idea of the context, so to work around this, we could make the current ``write`` a semi-private ``_write``, and create a separate method for the public API which collects all settings and calls ``_write``. ``Modflow6Simulation.write`` would then directly call ``_write``.

The only downside of this is that extending the WriteContext with some extra settings, might require some extra work, as arguments have to be added as well. However, I think that is worth it now, as letting users create internal objects to do what they want to do is also unwanted.

Something like:

```python

class Package:
...

    def write(
        self,
        pkgname: str,
        globaltimes: Union[list[np.datetime64], np.ndarray],
        directory: str,
        binary: bool = True,
        use_absolute_paths = False,
    ):
       write_context = WriteContext(directory, binary, use_absolute_paths)
       self._write(pkgname, globaltimes, write_context)

    @standard_log_decorator()
    def _write(
        self,
        pkgname: str,
        globaltimes: Union[list[np.datetime64], np.ndarray],
        write_context: WriteContext,
    ):
        directory = write_context.write_directory
        binary = write_context.use_binary
        self.write_blockfile(pkgname, globaltimes, write_context)

        if hasattr(self, "_grid_data"):
            if self._is_xy_data(self.dataset):
                pkgdirectory = directory / pkgname
                pkgdirectory.mkdir(exist_ok=True, parents=True)
                for varname, dtype in self._grid_data.items():
                    key = self._keyword_map.get(varname, varname)
                    da = self.dataset[varname]
                    if self._is_xy_data(da):
                        if binary:
                            path = pkgdirectory / f"{key}.bin"
                            self.write_binary_griddata(path, da, dtype)
                        else:
                            path = pkgdirectory / f"{key}.dat"
                            self.write_text_griddata(path, da, dtype)

### PR DESCRIPTION
**Pull request by JoerivanEngelen**
Fixes #1223

# Description
Remove WriteContext and ValidationContext from public API. 

# Checklist

- [x] Links to correct issue
- [ ] Unit tests were added

### DIFF
--- a/imod/mf6/model.py
+++ b/imod/mf6/model.py
@@ -305,7 +305,7 @@ class Modflow6Model(collections.UserDict, IModel, abc.ABC):
                 top, bottom, idomain = self.__get_domain_geometry()
                 k = self.__get_k()
                 mf6_hfb_pkg = merge_hfb_packages(mf6_hfb_ls, idomain, top, bottom, k)
-                mf6_hfb_pkg.write(
+                mf6_hfb_pkg._write(
                     pkgname=pkg_name,
                     globaltimes=globaltimes,
                     write_context=pkg_write_context,
--- a/imod/mf6/model.py
+++ b/imod/mf6/model.py
@@ -390,7 +390,7 @@ class Modflow6Model(collections.UserDict, IModel, abc.ABC):
             try:
                 if issubclass(type(pkg), GridAgnosticWell):
                     mf6_well_pkg = self._prepare_wel_for_mf6(pkg_name, validate_context)
-                    mf6_well_pkg.write(
+                    mf6_well_pkg._write(
                         pkgname=pkg_name,
                         globaltimes=globaltimes,
                         write_context=pkg_write_context,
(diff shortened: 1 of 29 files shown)

### CODE REVIEW

File: imod/mf6/model.py [hunk 1]
```diff
@@ -239,11 +239,55 @@ def validate(self, model_name: str = "") -> StatusInfoBase:
...
-        globaltimes,
+        modelname: str,
+        globaltimes: Union[list[np.datetime64], np.ndarray],
+        directory: str | Path,
+        use_binary: bool = True,
+        use_absolute_paths: bool = False,
+        validate: bool = True,
+    ):
+        """
+        Write MODFLOW6 model to file. Mote that this method is purely inteded
```
Manangka, line 252: inteded => intended

Manangka (approved):

PR #1293, opened 2024-11-15
Issue #1292 mete grids from imod5

### LINKED ISSUE #1292: MeteGrids:

JoerivanEngelen:
Fixes #1292

# Description
Context: MetaSWAP can map meteorological grids, which have a coarser grid but a very fine time resolution, to its svats. 
The ``mete_grid.inp`` file contains paths to meteorological information, stored in ASCII grids, and there are separate mapping files ``"svat2precgrid.inp"`` and ``"svat2etrefgrid.inp"``. Nota bene: MetaSWAP has hardcoded filenames for packages, hence why I can search for "mete_grid.inp". 

Adds the following:
- A ``MeteoGridCopy`` class to copy "mete_grid.inp" files, to avoid having to read and write the crazy amount of meteo files that exist in existing databases.
- A ``MeteoMapping.from_imod5_data`` method which looks for the first ascii grid in the ``mete_grid.inp`` file, and reads that to derive mappings from meteorological cell to metaswap grid cell.
- Add ``Imod5DataDict`` type alias, which can be implemented for other ``from_imod5_data`` methods in the future.

Only thing I doubt now is that I naively copy ``mete_grid.inp``, which works when the paths to ASCII grids are absolute, but not when they are relative. I don't think iMOD5 does this as well, so present databases all contain absolute paths. @HendrikKok do you foresee a usecase to support relative paths in ``MeteoGridCopy``?

# Checklist

- [x] Links to correct issue
- [x] Unit tests were added

### DIFF
--- a/imod/msw/meteo_mapping.py
+++ b/imod/msw/meteo_mapping.py
@@ -1,16 +1,47 @@
 from copy import deepcopy
+from pathlib import Path
 from typing import Any, Optional, TextIO
 
 import numpy as np
 import pandas as pd
 import xarray as xr
 
+import imod
 from imod.mf6.utilities.regrid import RegridderWeightsCache
 from imod.msw.fixed_format import VariableMetaData
 from imod.msw.pkgbase import MetaSwapPackage
 from imod.prepare import common
-from imod.typing import GridDataArray, IntArray
+from imod.typing import GridDataArray, GridDataDict, IntArray
 from imod.util.regrid_method_type import RegridMethodType
+
+
+def find_in_file_list(filename: str, paths: list[str]) -> str:
+    for file in paths:
+        if filename == Path(file[0]).name.lower():
+            return file[0]
+    raise ValueError(f"could not find {filename} in list of paths: {paths}")
+
+
+def open_first_meteo_grid(mete_grid_path: str, column_nr: int) -> xr.DataArray:
+    """
+    Find first meteo grid path in mete_grid.inp. Only read the first grid, so it
+    can be used to generate meteomappings.
+    """
+    if column_nr not in [2, 3]:
+        raise ValueError("Column nr should be 2 or 3")
+
+    f = open(mete_grid_path, "r")
+    lines = f.readlines()
+    meteo_filepath = Path(lines[0].split(",")[column_nr].replace('"', ""))
+    return imod.rasterio.open(meteo_filepath)
+
+
+def open_first_meteo_grid_from_imod5_data(
+    imod5_data: dict[str, GridDataDict], column_nr: int
+):
+    paths = imod5_data["extra"]["paths"]
+    metegrid_path = find_in_file_list("mete_grid.inp", paths)
+    return open_first_meteo_grid(metegrid_path, column_nr=column_nr)
 
 
 class MeteoMapping(MetaSwapPackage):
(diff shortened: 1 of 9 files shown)


imod/msw/utilities/common.py
```diff
@@ -1,6 +1,15 @@
...
 from imod.typing.grid import concat
 
 
 def concat_imod5(arg1: GridDataArray, arg2: GridDataArray) -> GridDataArray:
     return concat([arg1, arg2], dim="subunit").assign_coords(subunit=[0, 1])
+
+
+def find_in_file_list(filename: str, paths: list[str]) -> str:
+    for file in paths:
+        if filename == Path(file[0]).name.lower():
```
Manangka, line 13: What does the zeroth index of a file return?

imod/msw/meteo_mapping.py
```diff
@@ -1,18 +1,43 @@
...
+    Find first meteo grid path in mete_grid.inp. Only read the first grid, so it
+    can be used to generate meteomappings.
+    """
+    if column_nr not in [2, 3]:
+        raise ValueError("Column nr should be 2 or 3")
+
+    mete_grid_path = Path(mete_grid_path)
+
+    f = open(mete_grid_path, "r")
+    lines = f.readlines()
```
Manangka, line 30: Instead of readlines you can use readline to get the first line

Furthermore i wiould suggest using a context to make sure the file is closed when you're done reading it

```
with open(mete_grid_path, 'r') as file:
  first_line = file.readline()
  meteo_filepath = Path(first_line.split(",")[column_nr].replace('"', ""))
```

Manangka (approved):

imod/msw/meteo_mapping.py
JoerivanEngelen: Oomph good one! Missed the context when opening the file 😱

imod/tests/test_msw/test_evapotranspiration_mapping.py
```diff
@@ -134,62 +80,83 @@ def test_evapotranspiration_mapping_negative_dx(fixed_format_parser):
 
     assert_equal(results["svat"], np.array([1, 2, 3, 4, 5, 6]))
     assert_equal(results["row"], np.array([1, 2, 1, 2, 2, 2]))
-    assert_equal(results["column"], np.array([2, 2, 2, 2, 2, 1]))
+    assert_equal(results["column"], np.array([2, 2, 2, 2, 2, 3]))
```
HendrikKok, line 83: Was there a mistake in the old test?

HendrikKok: I guess because you moved the coordinates in the tests quite a bit it seems?

imod/msw/meteo_mapping.py
```diff
@@ -1,18 +1,43 @@
...
+    Find and open first meteo grid path in mete_grid.inp. This grid is enough to
+    generate meteomappings.
+    """
+    if column_nr not in [2, 3]:
+        raise ValueError("Column nr should be 2 or 3")
+
+    mete_grid_path = Path(mete_grid_path)
+
+    with open(mete_grid_path, "r") as f:
+        first_line = f.readline()
```
HendrikKok, line 30: MetaSWAP also supports constants as input (I found out recently). According to the manual there should be at least one meteo-file present. I would suggest a open_first_present_meteo_grid method.

HendrikKok (changes requested):
Nice work so far. I have one change request based on my recent experience.

In response to your question: I don't think we have to support relative paths in the mete_grid-file. Did not see this in practice.

imod/msw/meteo_mapping.py
JoerivanEngelen: Implemented!

HendrikKok (approved):

PR #1315, opened 2024-12-02
Issue #1313 recharge from imod5 cap data

### LINKED ISSUE #1313: Add ``imod.mf6.Recharge.from_imod5_cap_data``
Requirements
---------------

- Recharge cells where msw cells are active, with rate 0.0.
- Classmethod should be named ``from_imod5_cap_data``, consistent method added to ``LayeredWell``

JoerivanEngelen:
Fixes #1313

# Description
Construct an rch-package from iMOD5 data in the CAP package, loaded with the ``open_projectfile_data`` function. Package is
used to couple MODFLOW6 to MetaSWAP models. Active cells will have a recharge rate of 0.0.

At the moment, MetaSWAP can only be coupled to the first layer, as this is also the case for ``primod``. iMOD Coupler these days supports coupling to other layers as well, but ``primod`` doesn't. Picking this up for iMOD Python and ``primod`` is worthy a separate story.

In detail:
- Add ``Recharge.from_imod5_cap_data`` class method, to construct an empty Recharge package with 0.0 rate.
- Minor refactor: Put ``GridData`` helper functions to ``msw/utilities/imod5_converter.py``, so that they can be reused.

# Checklist

- [x] Links to correct issue
- [x] Unit tests were added

### DIFF
--- a/imod/mf6/model_gwf.py
+++ b/imod/mf6/model_gwf.py
@@ -330,6 +330,9 @@ class GroundwaterFlowModel(Modflow6Model):
                     wel_key, imod5_data, times
                 )
 
+        if "cap" in imod5_keys:
+            result["msw-rch"] = Recharge.from_imod5_cap_data(imod5_data)  # type: ignore
+
         # import ghb's
         imod5_keys = list(imod5_data.keys())
         ghb_keys = [key for key in imod5_keys if key[0:3] == "ghb"]
--- a/docs/api/changelog.rst
+++ b/docs/api/changelog.rst
@@ -14,6 +14,8 @@ Added
 
 - :class:`imod.msw.MeteoGridCopy` to copy existing `mete_grid.inp` files, so
   ASCII grids in large existing meteo databases do not have to be read.
+- :meth:`imod.mf6.Recharge.from_imod5_cap_data` to construct a recharge package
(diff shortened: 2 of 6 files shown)


imod/mf6/model_gwf.py
```diff
@@ -330,6 +330,9 @@ def from_imod5_data(
                     wel_key, imod5_data, times
                 )
 
+        if "cap" in imod5_keys:
+            result["msw-rch"] = Recharge.from_imod5_cap_data(imod5_data)  # type: ignore
```
Manangka, line 334: I would reconsider the place you put you converter code. The Recharge package now needs to now about the imod5 model. What would happen if we add another converter e.g. flopy? Then mf6 module also needs to know about that.

I would recommend moving all the converter code to a separate module. And in that module the code then knows about the imod5 data/packages and the mf6 data/packages

Manangka (changes requested):

imod/mf6/model_gwf.py
JoerivanEngelen: I agree, but it would be too large of a task to fix in this PR. I created this issue for this: https://github.com/Deltares/imod-python/issues/1319

Manangka (approved):

PR #1321, opened 2024-12-03
Issue #1320 Add CopyFiles package

### LINKED ISSUE #1320: Add ``CopyFiles`` MetaSWAP package, to copy settings and lookup tables from iMOD5 projectfiles
The iMOD5 projectfile contains an "EXTRA" section, with files to copy. These files usually contain settings and lookup tables for MetaSWAP, which have no direct relation to cell data. 

I've seen instances where mappings to meteo grids (``svat2precgrid.inp`` and ``svat2etrefgrid.inp``) are copied as well, but this is dangerous as it leads to errors if the grid changes. It also is not necessary for iMOD Python, as per https://github.com/Deltares/imod-python/pull/1315. 

Ideally, I would create MetaSwapPackage subclasses for each possible lookup table and settings file to read and write it back again, as it would allow users to parametrize MetaSWAP settings in Python. However, we don't have the necessary resources for that available and I don't see a big demand for it as is.

I think for the beta release, adding a simple ``CopyFiles`` package class is sufficient. Upon calling its ``.write`` method, it will copy files. 

Requirements
---------------
- Copy files listed under the "extra" section in the iMOD5 projectfile
- Does not write own .inp, instead, upon calling the ``.write`` method, files are copied to their destination
- Take list of paths, ideally store in xarray dataset, so MetaSWAP model can be dumped in the future.
- Do not copy the following files:
   * ``para_sim.inp``
   * ``mete_grid.inp``
   * ``svat2precgrid.inp``
   * ``svat2etrefgrid.inp``

JoerivanEngelen:
Fixes #1320

# Description
The iMOD5 projectfile contains an "EXTRA" section, with files to copy. These files usually contain settings and lookup tables for MetaSWAP, which have no direct relation to cell data.

This PR adds CopyFiles package to copy these files.

# Checklist
- [x] Links to correct issue
- [x] Unit tests were added


imod/tests/test_msw/test_copy_files.py
```diff
@@ -0,0 +1,69 @@
+from pytest_cases import parametrize_with_cases
+
+from imod.msw.copy_files import CopyFiles
+
+
+def setup_src_files(directory, filenames):
```
Manangka, line 6: You could call this create_test_files or write_test_files

imod/msw/copy_files.py
```diff
@@ -0,0 +1,55 @@
...
+from imod.logging.loglevel import LogLevel
+from imod.msw.pkgbase import MetaSwapPackage
+from imod.typing import Imod5DataDict
+
+_LOG_MESSAGE_TEMPLATE = """\
+Will not copy files {filtered}, these will be generated by iMOD Python
+instead."""
+
+
+class CopyFiles(MetaSwapPackage):
```
Manangka, line 18: A class name should be a Noun and its method should be verbs.
Consider naming this class FileCopier

Manangka, line 18: Why does this class inherit from the MetaSwapPackage?
It seems like a really general class. I don't know how it is being used but is it possible to use composition over inheritance? 

If it is possible to have this class without MetaSwapPackage as its parent class then you also don't need xarray as the data container and you have a member variable that is a list of paths

imod/msw/copy_files.py
```diff
@@ -0,0 +1,55 @@
...
+        super().__init__()
+        paths_da = xr.DataArray(
+            paths, coords={"file_nr": np.arange(len(paths))}, dims=("file_nr",)
+        )
+        self.dataset["paths"] = paths_da
+
+    @classmethod
+    def from_imod5_data(cls, imod5_data: Imod5DataDict):
+        paths = cast(list[list[str]], imod5_data["extra"]["paths"])
+        paths_unpacked = {Path(p[0]) for p in paths}
```
Manangka, line 29: Why are you using curly brackets and not block brackets? I thought curly brackets are only for dictionaries, but this seems to be a list.
The same thing occurs on line 39

imod/msw/copy_files.py
```diff
@@ -0,0 +1,55 @@
...
+        files_to_filter = (
+            "mete_grid.inp",
+            "para_sim.inp",
+            "svat2precgrid.inp",
+            "svat2etrefgrid.inp",
+        )
+        paths_filtered = [
+            str(p) for p in paths_unpacked if p.name.lower() not in files_to_filter
+        ]
+        paths_filtered_away = {str(p) for p in paths_unpacked} - set(paths_filtered)
```
Manangka, line 39: I would call this `paths_excluded` or `files_excluded` 
Likewse you could call `paths_filtered` `paths_included` or `files_included`

imod/msw/copy_files.py
JoerivanEngelen: This is a Python "set". They are very useful to check values in iterables for uniqueness. And if one set contains values not present, or present, in another set. For docs, see: https://docs.python.org/3/library/stdtypes.html#set

I wanted to make this more explicit by doing ``set([Path(p[0]) for p in paths])``, but Ruff disagreed with me. I guess the Ruff devs think the set is common knowledge as it is a Python builtin (though a bit more niche).

imod/msw/copy_files.py
JoerivanEngelen: I thought about that, my main reason why I decided to inherit MetaSwapPackage is that it makes the package object function consistently to other packages. For instance, we have plans to create a ``MetaSwapModel.dump`` method, which would dump all packages to file. To have a functional MetaSWAP model, we also need to store the paths to files that need to be copied somewhere. So by wrenching this list of strings/paths in a xarray Dataset, I'm quite sure no special-cased logic is required for this specific filecopy package and keeps things stored consistently in NetCDFs, instead of some custom textfile.

imod/tests/test_msw/test_copy_files.py
JoerivanEngelen: Good point, done!

imod/msw/copy_files.py
JoerivanEngelen: Done

imod/msw/copy_files.py
JoerivanEngelen: Done!

Manangka (approved):

PR #1356, opened 2024-12-19
Issue #1355 Slight optimization for performance regrid array

### LINKED ISSUE #1355: [PERFORMANCE] - During regrid array, full numpy arrays are loaded into memory for a simple check
Over here:
https://github.com/Deltares/imod-python/blob/54e0669424f9c5e375c6eb3f41ff33bebd8df17b/imod/mf6/utilities/regrid.py#L165

There is a call to ``da.values[()]`` which means arrays are fully loaded into memory just to check whether values are valid. Invalid values are always scalar, so the check needs to be able to shortcut with a call to ``is_scalar``:

```python
if is_scalar(da) and not _is_valid(da.values[()]):
```

In this way the check shortcuts on getting False from ``is_scalar``, saving a lot of loading data into memory.

JoerivanEngelen:
Fixes #1355

# Description
Slight performance improvement: Avoid checking if all arrays is_invalid. Only do when ``is_scalar``.

For test with 1e9 cells:

Before fix:
567 ms ± 26.1 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)
After fix:
545 ms ± 19.9 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)

Second time I ran these tests:
Before fix:
530 ms ± 21.6 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)
After fix:
503 ms ± 29.2 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)

Script to run:

```python
import imod
import numpy as np
from imod.mf6.utilities.regrid import _regrid_array, RegridderWeightsCache, RegridderType
import xarray as xr

dx = 1.
dy = -1.
xmin = 0.0
xmax = 10000.0
ymin = 0.0
ymax = 10000.0
layer = np.arange(10) + 1

da = xr.ones_like(imod.util.empty_3d(dx=dx, xmin=xmin, xmax=xmax, dy=dy, ymin=ymin, ymax=ymax, layer=layer))
target = imod.util.empty_2d(dx=10.0, xmin=xmin, xmax=xmax, dy=-10.0, ymin=ymin, ymax=ymax)

regridder_collection = RegridderWeightsCache()
regridder_name = RegridderType.OVERLAP 
regridder_function = "mean"

# In IPython, run:
# %timeit _regrid_array(da, regridder_collection, regridder_name, regridder_function, target)
```

# Checklist

- [x] Links to correct issue
- [ ] Unit tests were added

### DIFF
--- a/docs/api/changelog.rst
+++ b/docs/api/changelog.rst
@@ -52,6 +52,7 @@ Fixed
 - :func:`imod.evaluate.calculate_gxg`, upon providing a head dataarray chunked
   over time, will no longer error with ``ValueError: Object has inconsistent
   chunks along dimension bimonth. This can be fixed by calling unify_chunks().``
+- Improved performance of regridding package data.
 
 
 Changed
--- a/imod/mf6/utilities/regrid.py
+++ b/imod/mf6/utilities/regrid.py
@@ -168,7 +168,7 @@ def _regrid_array(
 
     # the dataarray might be a scalar. If it is, then it does not need regridding.
     if scalar_da:
-        return da.values[()]  # type: ignore [attr-defined]
+        return da.values[()]
 
     if isinstance(da, xr.DataArray):
         coords = da.coords
--- a/imod/mf6/utilities/regrid.py
+++ b/imod/mf6/utilities/regrid.py
@@ -162,11 +162,12 @@ def _regrid_array(
     """
 
     # skip regridding for scalar arrays with no valid values (such as "None")
-    if is_scalar(da) and not _is_valid(da.values[()]):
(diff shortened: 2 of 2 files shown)


Manangka (approved):
Would be interesting to see how well this scales for bigger datagrids

JoerivanEngelen:
> Would be interesting to see how well this scales for bigger datagrids

I don't foresee usecases for larger grids than 1e9 cells :P But I foresee use cases where a LOT of these grids have to be regridded. Regardless, it is going to at least slightly improve performance. This was some low-hanging fruit.

PR #1359, opened 2024-12-19
Issue #1103 Dump iMOD Python version number to toml

### LINKED ISSUE #1103: Add iMOD Python version number to dumped toml file
The ``Modflow6Simulation.dump`` method is at present a method to temporarily store Modflow 6 simulations inbetween scripts. For the version 1.0, we'd like this to be a replacement of the iMOD 5 databases, meaning at minimum we need to add the iMOD Python version number to the toml file. 

Given that we expected that iMOD Python still is quite under some development after version 1.0, we need to do one of the following upon reading a toml with an older version:

1. Throw warning that things might not work given that this is an older version.
2. Throw error to force users to update their database.

I think starting with option 1 is the easiest to inform users. 2 requires more work, but might be a hard requirement when we make some very significant changes.

Upon reading a toml file with a newer version than the current iMOD Python, we can throw an error ushering people to update their iMOD Python. Just like pixi does.

JoerivanEngelen:
Fixes #1103

# Description
Dump iMOD Python version number to toml, throw ``UserWarning`` if version number mismatch upon loading dumped data or when no version number could be found.

Also remove python version check for pytest as we are not going below 3.7
 
# Checklist
- [x] Links to correct issue
- [x] Unit tests were added

### DIFF
--- a/imod/mf6/simulation.py
+++ b/imod/mf6/simulation.py
@@ -871,6 +871,9 @@ class Modflow6Simulation(collections.UserDict, ISimulation):
         directory.mkdir(parents=True, exist_ok=True)
 
         toml_content: DefaultDict[str, dict] = collections.defaultdict(dict)
+        # Dump version number
+        toml_content["version"] = {"imod-python": imod.__version__}
+        # Dump models and exchanges
         for key, value in self.items():
             cls_name = type(value).__name__
             if isinstance(value, Modflow6Model):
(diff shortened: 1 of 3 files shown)


imod/mf6/simulation.py
```diff
@@ -87,6 +87,22 @@ def get_packages(simulation: Modflow6Simulation) -> dict[str, Package]:
     }
 
 
+def validate_version(version_saved: str) -> None:
+    if version_saved and version_saved["imod-python"] != imod.__version__:
+        warnings.warn(
+            "iMOD Python version in current environment different from version with which simulation was dumped."
```
Manangka, line 93: The iMOD Python version in the current environment is different from the version with which the simulation was dumped.

Manangka (approved):
Alright. It's good to save the version number. This opens some door.

For instance, In the future we can write 'migrators' to upgrade old versions to new ones.

I'm not sure if giving a warning when the version number is different then the current version. A warning is only relevant if the versions are incompatible (or should it then be an error?). You could log the mismatch of the version as an info message in the logger

JoerivanEngelen:
I updated the code to log to inform users on used versions without throwing all kinds of warnings. Warning upon version mismatch was indeed overboard.

PR #1369, opened 2025-01-03
Issue #1368 Add Multidoc utiity for creating versioned documentation

### LINKED ISSUE #1368: Add multi-version documentation

Manangka:
Fixes #1368

# Description
This is the first step to having multi-version documentation. The steps needed are:

1) Create multi-version documentation
2) Clear the gh-pages branch. Add manually the master documentation to gh-pages and disable the deploy documentation pipeline
3) Update pipeline to update master documentation each time a PR is merged to master
5) Refactor "deploy documentation" to create and add the deployed version documentation
6) Add older versions to the documentation
7) Add helper pipeline to manually remove versions from the documentation

This commit introduces multidoc.py. This is a small helper script that helps build the multiversion documentation.
It has 3 commands:
- add-version
- update-version
- remove-version

When a version is added or updated the version is checkout using the git worktree command.
This creates a separate folder which contains the branch/tag of which we want to build the documentation.
Building the documentation in the worktree is done in the same way as we do it currently

To have multi-version support 2 changes need to be made. In the conf.py file we need to add `switcher` to the `html_options`. Furthermore we need to create a switcher.json file which contains all the available version in the documentation.

Since older version don't have the updated  `html_options`  a patch file is included. When calling `add-version` or `update-version` this patch is appllied. This way we can also build documentation for older version.

The `add-version` and `update-version` both build the documentation. The `add-version` also updates the `switcher.json` file to contain the new version.

The `remove-version` command removes the specified version from the documentation and updated the `switcher.json` accordingly

There are some optional flags that can be used.
A useful one in the `--local-build` flag. This flag rewrites the documentation url to a local filesystem url. This way you can use the switcher also locally. 
When running it locally I did a CORS error from chrome. This can be circumenvented by running"
`chrome.exe --user-data-dir="C:/Chrome dev session" --disable-web-security`

# Checklist

- [ ] Links to correct issue
- [ ] Unit tests were added


docs/conf.py
```diff
@@ -126,6 +127,20 @@
...
+    capture_output=True,
+    shell=True,
+    text=True,
+).stdout.strip()
+
+env = os.environ
+json_url = (
+    "https://deltares.github.io/imod-python/_static/switcher.json"
+    if "JSON_URL" not in env
+    else env["JSON_URL"]
```
JoerivanEngelen, line 141: Please add a comment when JSON_URL is added to the environment when running multidoc.py

JoerivanEngelen (commented):
Thanks for the efforts, looks good. My main thoughts:

- I'm a bit surprised this required quite some custom code: that there is not a plug and play answer for our usecase. Did you take inspiration from any repositories? Might be good to add here, so we can check how they updated their code whenever this breaks after an update.
- I thought we set up pixi tasks so that all tasks related to our builds are called from pixi. But now we have a python script to run around pixi again?
- This adds some extra complexity to our builds, given the extra custom code. This is because we don't want to rebuild docs for every version, if I'm not mistaken. I see no description online why we are doing this. Some requirements would help here. Could you add that to the issue description?
- The developer docs need to be extended how to run this. Probably you skipped on that in awaiting of my review.

PR #1377, opened 2025-01-10
Issue #1376 active idomain larger than zero

### LINKED ISSUE #1376: Find active cells with idomain > 0 instead of idomain == 1
The LHM apparently has > 1000 cells where IDOMAIN == 2. I noticed that wells located in these cells get masked away. For both iMOD5 as well as MODFLOW 6, IDOMAIN > 0 means active cell. iMOD Python, however, enforces/assumes IDOMAIN == 1, but we don't actually validate this.

I think it is more user friendly to support IDOMAIN > 0. A quick search showed that we only have a few files where active is computed as idomain == 1, so it appears to be low hanging fruit.

JoerivanEngelen:
Fixes #1376 

# Description
This aligns the way we find active cells, consitently with MODFLOW6: idomain > 0 is now the active domain, instead of idomain == 1. Also I tried my best to update all docstrings.

# Checklist

- [x] Links to correct issue
- [x] Unit tests were added

### DIFF
--- a/imod/mf6/out/dis.py
+++ b/imod/mf6/out/dis.py
@@ -325,7 +325,7 @@ def dis_indices(
     * the right neighbor is +(1)
     * the front neighbor is +(number of cells in a column)
     * the lower neighbor is +(number of cells in a layer)
-    * lower "pass-through" cells (idomain == -1) are multitude of (number of
+    * lower "pass-through" cells (idomain <0) are multitude of (number of
       cells in a layer)
 
     Parameters
--- a/imod/tests/fixtures/backward_compatibility_fixture.py
+++ b/imod/tests/fixtures/backward_compatibility_fixture.py
@@ -21,7 +21,7 @@ def imod5_dataset():
 
     # Fix data for ibound  as it contains floating values like 0.34, 0.25 etc.
     ibound = data["bnd"]["ibound"]
-    ibound = ibound.where(ibound <= 0, 1)
+    ibound = ibound.where(ibound <= 0, 2)
     data["bnd"]["ibound"] = ibound
     return data, pd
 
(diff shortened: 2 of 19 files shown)


imod/mf6/out/dis.py
```diff
@@ -325,7 +325,7 @@ def dis_indices(
     * the right neighbor is +(1)
     * the front neighbor is +(number of cells in a column)
     * the lower neighbor is +(number of cells in a layer)
-    * lower "pass-through" cells (idomain == -1) are multitude of (number of
+    * lower "pass-through" cells (idomain <0) are multitude of (number of
```
Manangka, line 328: We have different definitions for idomain <0. I saw in another file that we defined is as `vertical passthrough`. Here is its defined as `pass-through`. We should pick 1 way of writing it and apply that consistently

imod/tests/fixtures/backward_compatibility_fixture.py
```diff
@@ -21,7 +21,7 @@ def imod5_dataset():
 
     # Fix data for ibound  as it contains floating values like 0.34, 0.25 etc.
     ibound = data["bnd"]["ibound"]
-    ibound = ibound.where(ibound <= 0, 1)
+    ibound = ibound.where(ibound <= 0, 2)
```
Manangka, line 24: Just wondering, when will the idomain ever be 2? Does it have a special meaning?

Manangka (approved):

imod/tests/fixtures/backward_compatibility_fixture.py
JoerivanEngelen: For MODFLOW6 it has no special meaning. Users can give it their own special meaning. For example, the LHM sets IDOMAIN 2 in areas in Germany where a lot of groundwater is pumped for coal mines. This is just so that users can use this to select data.

imod/mf6/out/dis.py
JoerivanEngelen: Done, went for "vertical passthrough"

PR #1380, opened 2025-01-14
Issue #1379 clip wel timeseries before resampling

### LINKED ISSUE #1379: iMOD5 timesteps in well dataset preceding model times should be ignored in timeseries resampling
**Bug description**
I ran into an issue with the LHM where resampling was not done as expected.
Say the well timeseries in the iMOD5 database is as follows:

```
# time     rate
1899-01-01 0.0
2010-01-01 350.0
2010-03-01 350.0
```

And the simulation times we want to import the model for ``2010-01-02``, we expect a rate of 350 m3/d. However the time weighted averaging (required for upsampling timeseries) includes the hundred years of with 0.0 m3/d, so it returns something like 0.6 m3/d. From the well timesteps preceding the first simulation timestep ``2010-01-02``, only the last (in this case ``2010-01-01``) needs to be kept, the rest can be ignored.

JoerivanEngelen:
Fixes #1379

# Description
Fix bug in ``resample_timeseries`` where all timesteps in the iMOD5 well database preceding simulation ``times`` were included in the time weighted averaging of the first timestep. Only the last timestep preceding ``times`` should be included, the rest should be ignored/discarded.

# Checklist

- [x] Links to correct issue
- [x] Unit tests were added

### DIFF
--- a/imod/util/expand_repetitions.py
+++ b/imod/util/expand_repetitions.py
@@ -56,7 +56,7 @@ def resample_timeseries(
     Parameters
     ----------
     well_rate:  pd.DataFrame
-        input timeseries for well
+        input timeseries for a single well
     times: datetime
         list of times on which the output dataframe should have entries.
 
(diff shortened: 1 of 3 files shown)


imod/util/expand_repetitions.py
```diff
@@ -75,6 +75,13 @@ def resample_timeseries(
         on="time",
         validate="m:m",
     ).fillna(method="ffill")
+    # Catch case where multiple well rates precede first timestep on multiple
+    # occasions, these have to be clipped to avoid including them in the
+    # computation of weighted averages later in the function. We only want to
+    # include one well timestep before the first output timestep at maximum.
```
Manangka, line 81: You mention that we only want to include one well timestep before the first output timestep. Which one should that be. The oldest? The latest? Should we mention it in the comment or is it of no importance?

Manangka (approved):

PR #1396, opened 2025-01-29
Issue #1395 Change default value for missing layers for angle1

### LINKED ISSUE #1395: Incorrect value used to fill missing layers in angle1 array in NPF package
**Bug description**
Users do not have to set values for each layer for the ANI package in the projectfile. We have utilities to fill these missing layers with a constant value. Right now we use the incorrect value to fill missing layers: Should be 90, is 0 at the moment.

JoerivanEngelen:
Fixes #1395

# Description
Use proper value for missing layers in projectfile.

# Checklist

- [x] Links to correct issue
- [x] Unit tests were added

### DIFF
--- a/imod/mf6/npf.py
+++ b/imod/mf6/npf.py
@@ -510,7 +510,7 @@ class NodePropertyFlow(Package, IRegridPackage):
                 angle1 = imod5_data["ani"]["angle"]
                 angle1 = 90.0 - angle1
                 angle1 = xr.where(angle1 < 0, 360.0 + angle1, angle1)
-                angle1 = fill_missing_layers(angle1, target_grid, 0)
+                angle1 = fill_missing_layers(angle1, target_grid, 90)
                 data["angle1"] = angle1
 
         icelltype = zeros_like(target_grid, dtype=int)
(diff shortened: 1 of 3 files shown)


Huite:
See my comment in the issue: entering a value for ANGLE will likely slow MF6 down. Better to apply the trigonometry yourself.

imod/mf6/npf.py
```diff
@@ -510,7 +510,7 @@ def from_imod5_data(
                 angle1 = imod5_data["ani"]["angle"]
                 angle1 = 90.0 - angle1
                 angle1 = xr.where(angle1 < 0, 360.0 + angle1, angle1)
-                angle1 = fill_missing_layers(angle1, target_grid, 0)
+                angle1 = fill_missing_layers(angle1, target_grid, 90)
```
Manangka, line 513: Should be 90.0 to be consistent with the way other angles are written

Manangka (approved):

imod/mf6/npf.py
JoerivanEngelen: Nice catch, fixed!

JoerivanEngelen:
> See my comment in the issue: entering a value for ANGLE will likely slow MF6 down. Better to apply the trigonometry yourself.

Created a separate issue for this https://github.com/Deltares/imod-python/issues/1397

Issue #1331, opened 2024-12-11
Python 3.13: Remove use of Path as contexts


JoerivanEngelen:
We are using Path objects as contexts in our code base, [which does nothing since Python 3.9](https://docs.python.org/3/whatsnew/3.13.html#id4).

In Python 3.13 this will throw an error:

![Image](https://github.com/user-attachments/assets/273ae859-079a-4d8a-8435-bad9a5789aa5)

PR #1404, opened 2025-02-04
Add python 3.13 support

### LINKED ISSUE #1317: [Bug] - iMOD Python breaks upon import with Python 3.13
**Bug description**
Python 3.13 has been released since 1 month. I got a report from that iMOD Python broke import with Python 3.13.

**Steps to reproduce**

1. What error message do you get?

Image

4. Can you create a minimum reproducible example?
   ```python
   # script.py
   import imod
   ```

**Desktop (please complete the following information):**
 - OS: [e.g. Windows 11]
 - iMOD Python Version [e.g. 0.18.0]

**Additional context** 
Add any other context about the problem here. The more information you can
provide us, the quicker we respond and fix the problem.

Refinement
-------------

- [x] Wait for numba release which supports python 3.13
- [ ] #1331
- [ ] Update TeamCity pipeline to also include python 3.13
- [x] #1332

Manangka:
Fixes #1317 and #1331

# Description
With `Numba` now also supporting python 3.13 we can also add 3.13 support to iMOD-python.
This PR adds a 3.13 environment to the pixi.toml and also adds an import test of it to the TC pipelines.

One line of code had to be rwritten to be 3.13 compatible. The `importlib.resources.files` no longer provides a context manager and has to wrapped using the `importlib.resources.as_file` method

# Checklist

- [ ] Links to correct issue
- [ ] Unit tests were added

### DIFF
--- a/.teamcity/Templates/PipPythonTemplate.kt
+++ b/.teamcity/Templates/PipPythonTemplate.kt
@@ -45,7 +45,8 @@ object PipPythonTemplate : Template({
                 "python_env", listOf(
                     value("py310", label = "python 3.10"),
                     value("py311", label = "python 3.11"),
-                    value("py312", label = "python 3.12")
+                    value("py312", label = "python 3.12"),
+                    value("py313", label = "python 3.13")
                 )
             )
         }
--- a/docs/api/changelog.rst
+++ b/docs/api/changelog.rst
@@ -12,6 +12,7 @@ The format is based on `Keep a Changelog`_, and this project adheres to
 Added
 ~~~~~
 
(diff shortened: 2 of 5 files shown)


JoerivanEngelen (approved):
Thanks for the fix! Could you update this in the changelog as well?

"Fixed an ImportError which was thrown upon calling ``import imod`` with Python 3.13"

I think we are still waiting for this one to succeed: https://github.com/conda-forge/numba-feedstock/pull/152 so we can really state that we are supporting Python 3.13. We can then update the documentation (see: "Which python?" section of installation guide).

PR #1426, opened 2025-02-13
Issue #1414 add cleanup hfb

### LINKED ISSUE #1414: Expose functionality to clip HorizontalFlowBarrier class
We have quite some internal logic to clip HorizontalFlowBarrier packages away from inactive cells. However, it isn't really exposed to the user. Right now we throw validation errors when barriers are intersecting a grid. When users set strict_hfb_validation to false, the models run without error as barrier segments connecting to inactive cells are filtered away. The latter is hard to do when running models with ``primod`` (probably worthy of its own issue).

I see two ways we could do this. Either by modifying the already existing ``mask`` method, which at present doesn't do anything because vector data is independent from grid. I therefore is quite a useless method right now, just existing to keep the API consistent. The other way we could go by this is introduce a ``cleanup_hfb`` function, which takes a grid an cleans it based on that. Or both?

JoerivanEngelen:
Fixes #1414 

# Description
- Adds ``cleanup_hfb`` function to clip barrier lines/Z polygons outside the active domain. This unfortunately cannot deal with the edge case where a hfb line is snapped to the outer edge of the active domain. (These edges are filtered out when calling ``to_mf6_pkg``, but a ValidationError is thrown when ``strict_hfb_validation`` is switched on).
- Adds ``HorizontalFlowBarrierBase.cleanup`` method
- Moves some utility functions from ``imod.mf6.utilities`` to ``imod.prepare.hfb`` and ``imod.util.hfb``, to avoid circular imports.

# Checklist

- [x] Links to correct issue
- [x] Unit tests were added


imod/prepare/hfb.py
```diff
@@ -217,3 +222,55 @@ def linestring_to_trapezoid_zpolygons(
         _line_to_trapezoid_zpolygon(x, y, zt, zb)
         for x, y, zt, zb in zip(x_pairs, y_pairs, zt_pairs, zb_pairs)
     ]
+
+
+def _clip_linestring(
```
Manangka, line 227: Why did you move these method to this class?
The original intent of moving them to a separate clip class was to keep the package classes small while keeping the similar functionality grouped.  Which also makes it easier to unit test. This was done using the Visitor Pattern.

I wouldn't move this to the hfb class because we now get a mix of clip methods being in a separate clip class and clip methods being in classes specificly related to a certain package class

Manangka (changes requested):

imod/prepare/hfb.py
JoerivanEngelen: The main reason is that I ran into circular imports, as the logic was put in the ``imod/mf6/utilities`` namespace. The cleanup was in the ``imod/prepare/cleanup`` namespace, which is imported in the ``imod/mf6/hfb.py`` namespace.

So probably what is required is moving ``imod/mf6/utilties/clip.py`` to ``imod/util/clip.py``. This, however heavility depending on ``imod/mf6/interfaces``, so these probably have to be moved one level up as well. That is probably better.

imod/prepare/cleanup.py
```diff
@@ -336,3 +337,31 @@ def cleanup_wel(
         cleaned_wells, minimum_thickness
     )
     return cleaned_wells
+
+
+def cleanup_hfb(
```
Manangka, line 342: This could also be called `clean_linedata` or `remove_inactive_hfb_data`

Manangka (approved):

imod/prepare/cleanup.py
JoerivanEngelen: This is public API, so I think I keep this to the MODFLOW6 package names, which is what users are familiar with.

PR #1446, opened 2025-02-21
Issue #1445 fix performance big model

### LINKED ISSUE #1445: [Bug] - Poor performance deriving dummy MODFLOW6 package for MetaSWAP coupling for big model
**Bug description**
@WouterSwierstra tried reading a large iMOD5 model, if I recall correctly: (30, 26, 1200, 1200), and writing it was extremely slow: Over 30 minutes. Just inferring from files, it appeared to be the MetaSWAP recharge package. I asked Wouter to profile it and just analysed the data. The profiler mentioned 87% of the time was spent in this function https://github.com/Deltares/imod-python/blob/a5f0cd35100106283f7353be179102474536c1b0/imod/mf6/wel.py#L531 . So I think what happened: The msw-sprinkling.wel file was written, but we were still waiting for all the corresponding data files to be written, before the msw-rch.rch could be written. Reproducing it should be quite simple: Just creating some large CAP data grids and seeing what takes very long. Also performance might be worse if loaded as dask. 

See Snakeviz table:

Image

JoerivanEngelen:
Fixes #1445

It turned out this performance issue pertained from a performance issue where calling ``xr.DataArray().isin(pd.Series())`` was very slow for large datasets, whereas ``pd.Series().isin(pd.Series())`` was a lot faster. 

# Description
- Fix performance issue when writing large MODFLOW6 WEL packages
- Add test which tests this performance, I had to add a small pytest plugin for this.

# Checklist
- [x] Links to correct issue
- [x] Unit tests were added

### DIFF
--- a/imod/mf6/wel.py
+++ b/imod/mf6/wel.py
@@ -536,8 +536,8 @@ class GridAgnosticWell(BoundaryCondition, IPointDataPackage, abc.ABC):
             filtered_ids = well_data_filtered["id"].to_dataframe()["id"]
         else:
             filtered_ids = well_data_filtered["id"]
-        is_missing_id = ~filtered_ids.isin(well_data["id"])
-        return filtered_ids[is_missing_id]
+        is_missing_id = ~well_data["id"].isin(filtered_ids.unique())
+        return well_data["id"].loc[is_missing_id].unique()
 
     def to_mf6_package_information(
         self, filtered_wells: list[str], reason_text: str
--- a/imod/mf6/wel.py
+++ b/imod/mf6/wel.py
@@ -531,12 +531,13 @@ class GridAgnosticWell(BoundaryCondition, IPointDataPackage, abc.ABC):
     def gather_filtered_well_ids(
         self, well_data_filtered: pd.DataFrame | xr.Dataset, well_data: pd.DataFrame
     ) -> list[str]:
-        filtered_well_ids = [
-            id
-            for id in well_data["id"].unique()
-            if id not in well_data_filtered["id"].values
-        ]
-        return filtered_well_ids
+        # Work around performance issue with xarray isin for large datasets.
+        if isinstance(well_data_filtered, xr.Dataset):
+            filtered_ids = well_data_filtered["id"].to_dataframe()["id"]
+        else:
+            filtered_ids = well_data_filtered["id"]
+        is_missing_id = ~filtered_ids.isin(well_data["id"])
+        return filtered_ids[is_missing_id]
 
     def to_mf6_package_information(
         self, filtered_wells: list[str], reason_text: str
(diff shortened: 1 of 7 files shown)


imod/mf6/wel.py
```diff
@@ -531,12 +531,13 @@ def _to_mf6_pkg(
...
         self, well_data_filtered: pd.DataFrame | xr.Dataset, well_data: pd.DataFrame
     ) -> list[str]:
-        filtered_well_ids = [
-            id
-            for id in well_data["id"].unique()
-            if id not in well_data_filtered["id"].values
-        ]
-        return filtered_well_ids
+        # Work around performance issue with xarray isin for large datasets.
+        if isinstance(well_data_filtered, xr.Dataset):
```
Manangka, line 535: Do you need to convert it to a dataframe? Can't you do directly access the values by `well_data_filtered["id"].values` ?

Or if it needs to be a dataframe wouldn't it better to construct it like 
`pd.DataFrame(well_data_filtered["id"].values)`?

Huite:
```python
        filtered_well_ids = [
            id
            for id in well_data["id"].unique()
            if id not in well_data_filtered["id"].values
        ]
```

https://www.tumblr.com/accidentallyquadratic

?

pixi.toml
```diff
@@ -128,6 +128,7 @@ vtk = { version = ">=9.0", build = "*qt*" }
 xarray = ">=2023.08.0"
 xugrid = ">=0.11.0"
 zarr = "*"
+pytest-timeout = ">=2.3.1,<3"
```
Manangka, line 131: What does this do?

Manangka (approved):

JoerivanEngelen:
> ```python
>         filtered_well_ids = [
>             id
>             for id in well_data["id"].unique()
>             if id not in well_data_filtered["id"].values
>         ]
> ```
> 
> https://www.tumblr.com/accidentallyquadratic
> 
> ?

Yeah I first thought this was the issue, but it turned out the conversion to a pandas series was essential to get proper performance.  Calling xarray.DataArray.isin(pd.Series) was just as slow as the original solution.

pixi.toml
JoerivanEngelen: Sorry, I should have elaborated more in the description. This pytest plugin allows you to set a timeout on tests. If changes will result on this test becoming slow again, it will show. This avoids having unittests that might run for an hour and then let TeamCity crash, after which the problems are less clear.

imod/mf6/wel.py
JoerivanEngelen: The function now accepts both xr.Dataset as well as pd.Series. It is called twice in the ``_to_mf6_pkg`` method, once with a pd.Series, once with a xr.Dataset.

The problem with using ``.values`` is that there is a difference between xarray DataArrays and pandas Series: xarray has a ``.values`` method, whereas pandas has a ``values`` property. I therefore thought calling ``to_dataframe`` made it more explicit that were are converting an xarray DataArray to a pandas Dataframe.

PR #1450, opened 2025-02-24
Issue #1436 work around dask issue

### LINKED ISSUE #1436: [Bug] - Derivation MetaSWAP mappings fails with Dask >= 2025.2.0
**Bug description**

By sheer coincidence, @WouterSwierstra and I ran into a bug that arose with the latest Dask release. Problematic are lines like this:

https://github.com/Deltares/imod-python/blob/4c4e6e13c0f7f15dd98dd482655ad819296c0231/imod/msw/coupler_mapping.py#L120

If we load this into memory as numpy arrays, everything is fine, however keeping them as lazy objects causes values not to be set or causes an error like: ``ValueError: cannot broadcast shape (10,) to shape (nan,)``. I think I'm doing something here that was not intended to work with Dask and somehow magically worked. 

And even worse:

https://github.com/Deltares/imod-python/blob/4c4e6e13c0f7f15dd98dd482655ad819296c0231/imod/msw/grid_data.py#L98

Here it tries to set a read-only array in case Dask is used: ``.values`` then returns a read-only array. 

Related issue: https://github.com/dask/dask/issues/11753, this provides a potential solution:

> In the meantime, dask-ml should use da.where instead (which is what da.Array.__setitem__ calls internally) unless there is a use case about applying the same functions to numpy arrays, which would take a performance hit?

So I therefore think using a ``where`` is a lot safer here. Operations are on 2D grids, without time dimension, so I don't foresee huge performance issues by doing this.

Furthermore: The fact that we missed this is probably caused by our test bench loading everything into memory, otherwise this bug would have surfaced earlier. Dask 2025.2.0 is in the present dev environment.

Refinement
-------------

- [ ] In the MetaSWAP mapping derivations that use fancy indexing, use ``where`` instead. 
- [ ] Add tests where data is not loaded into memory and run with dask.

JoerivanEngelen:
Fixes #1436 

# Description
Work around Dask issue by forcing a load of idomain and svat into memory. These grids will never be as large as the boundary condition grids with a time axis, so I think it is acceptable. At least for the LHM it is. I couldn't get it to work with a ``where``: numpy only accepts 1D arrays on a boolean indexed 3D as long as the 1D array has the length of one of the dimensions, it then does what we want to do here. However when this is not the case, which is 95% of the time, numpy throws an error. I therefore had to resort to loading arrays into memory.

Furthermore I refactored ``test_coupler_mapping.py`` to be able to reduce some code duplication and add a test case with a dask array.

Furthermore

# Checklist
- [x] Links to correct issue
- [x] Unit tests were added

### DIFF
--- a/imod/msw/grid_data.py
+++ b/imod/msw/grid_data.py
@@ -91,11 +91,14 @@ class GridData(MetaSwapPackage, IRegridPackage):
         active = self.dataset["active"]
 
         isactive = area.where(active).notnull()
-
-        index = isactive.values.ravel()
-
         svat = xr.full_like(area, fill_value=0, dtype=np.int64).rename("svat")
-        svat.values[isactive.values] = np.arange(1, index.sum() + 1)
+        # Load into memory to avoid dask issue
+        # https://github.com/dask/dask/issues/11753
+        isactive.load()
+        svat.load()
+        
+        index = isactive.values.ravel()
+        svat.data[isactive.data] = np.arange(1, index.sum() + 1)
 
         return index, svat
 
--- a/imod/msw/grid_data.py
+++ b/imod/msw/grid_data.py
@@ -96,7 +96,7 @@ class GridData(MetaSwapPackage, IRegridPackage):
         # https://github.com/dask/dask/issues/11753
         isactive.load()
         svat.load()
-        
+
         index = isactive.values.ravel()
         svat.data[isactive.data] = np.arange(1, index.sum() + 1)
 
--- a/docs/api/changelog.rst
+++ b/docs/api/changelog.rst
@@ -93,10 +93,13 @@ Fixed
   written with ``validate`` set to True.
 - Fixed part of the code that made Pandas, Geopandas, and xarray throw a lot of
   ``FutureWarning``s and ``DeprecationWarning``s.
-- Fixed performance issue when coverting very large wells (>10k) with
+- Fixed performance issue when converting very large wells (>10k) with
   :meth:`imod.mf6.Well.to_mf6_pkg` and :meth:`imod.mf6.LayeredWell.to_mf6_pkg`,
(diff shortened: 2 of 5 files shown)


imod/msw/grid_data.py
```diff
@@ -91,11 +91,14 @@ def generate_index_array(self):
         active = self.dataset["active"]
 
         isactive = area.where(active).notnull()
+        svat = xr.full_like(area, fill_value=0, dtype=np.int64).rename("svat")
```
Manangka, line 94: I don't understand how this fixes the issue.
I looked at the github issue you linked to.
There the problematic statement seems to be `a[mask] = a[mask]`. 
In this case is that `svat.data[isactive.data] = np.arange(1, index.sum() + 1)`?

It seems that the bug is being recognized.
Wouldn't it be better to pin dask to an older version until this issue is resolved?

JoerivanEngelen: Dask has slightly different behavior here than numpy. I think we are threading into more obscure features of numpy, of which I'm not certain they are in the scope of Dask or not.

From what I read, the dask developer was surprised ``a[mask] = b`` worked in the first place. Which is what we do here. Furthermore, it seems 2025.1.0 also shows some weird behavior with masks. I therefore think it is safer to load these arrays into memory (turning them into regular numpy arrays) for now. That is less hassle then pinning to older versions, as that entails patching the conda-forge repopatches feedstock etc.

JoerivanEngelen: Just looked further into this, I think our issue is unrelated to that issue. So what I did never worked with dask in the first place.

```
ValueError: Boolean index assignment in Dask expects equally shaped arrays.
Example: da1[da2] = da3 where da1.shape == (4,), da2.shape == (4,) and da3.shape == (4,).
Alternatively, you can use the extended API that supportsindexing with tuples.
Example: da1[(da2,)] = da3.
```

Manangka (approved):

Issue #1410, opened 2025-02-06
Tracking issue: Peelvenen MetaSWAP test


JoerivanEngelen:
I sat together with Wouter yesterday to convert the Peelvenen iMOD5 model to a coupled MODFLOW6-MetaSWAP-Ribasim model. We ran into some issues, but could work around them and the model ran without error. There were still some issues with the generated MetaSWAP model input though. This issue is to track progress on resolving these issues and the hiccups we ran into.

PR #1452, opened 2025-02-24
Don't overwrite budget terms, sum instead

### LINKED ISSUE #1451: [Bug] - ``open_cbc``: Multiple entries per cell in a single package are not summed
**Bug description**

iMOD Python never generates boundary condition packages with multiple entries per cell. MODFLOW6, however, supports this. Investigating #1443, I found that iMOD5 quite often generates a river package with two entries for a single cell, to convert its infiltration factors. Flopy can also generate multiple entries for a single cell, we've seen some of these models in the wild. If a user starts to use iMOD Python to read a CBC file from a model generated with Flopy or iMOD5, we cannot guarantee that the fluxes now read are correct. iMOD Python probably takes the value of the last entry in a single cell.

Things we could do to ``imod.mf6.out.open_cbc``:

1. Throw an error if cellids in a CBC file are not unique
2. Groupby cellids and timestep, then sum
3. Expand dimensions along a new ``n_entry`` dimension, of which the size has to be found in some way.

Huite:
Sum the budget terms instead of having them overwite each other in case of duplicate indices values.

Fixes #1451

### DIFF
--- a/imod/mf6/out/cbc.py
+++ b/imod/mf6/out/cbc.py
@@ -304,5 +304,8 @@ def read_imeth6_budgets_dense(
     table = read_imeth6_budgets(cbc_path, count, dtype, pos)
     if indices is None:
         indices = table["id1"] - 1  # Convert to 0 based index
-    out[indices] = table[return_variable]
+    # Zero the relevant values, overwrite the NaN value.
+    out[indices] = 0
+    # Sum all the budget terms.
+    np.add.at(out, indices, table[return_variable])
     return out.reshape(shape)


JoerivanEngelen (approved):

JoerivanEngelen:
Thanks for fixing this! Very simple changeset! I added a unittest to verify it worked and also tested it with the example provided by  @WouterSwierstra in https://github.com/Deltares/imod-python/issues/1410?issue=Deltares%7Cimod-python%7C1443, where it worked.

PR #1460, opened 2025-02-26
Issue #1057 remove duplicate entries CouplerMapping

### LINKED ISSUE #1057: Sprinkling from first model layer leads to wrong id's in mod2svat.inp
The file mod2svat.inp should only contains unique sets of id's (mod_id, svat, layer). Sprinkling from the first model layer however, results in a doubled set of id's. This leads to unwanted behavior in MetaSWAP. 

The _create_well_id method should exclude all id's associated with the first model layer.

JoerivanEngelen:
Fixes #1057

# Description
Remove duplicate entries in ``mod2svat.inp``. The issue mentions only layer 1, but at some point in the future it might be possible to couple MetaSWAP recharge to other layers as well. Our unittest cases already had duplicate entries (which MetaSWAP would throw an error upon if we ever tried using them), so I only had to update these.

# Checklist
- [x] Links to correct issue
- [ ] Unit tests were added


imod/msw/coupler_mapping.py
```diff
@@ -97,6 +97,10 @@ def _render(
         dataframe = pd.DataFrame(
             data=data_dict, columns=list(self._metadata_dict.keys())
         )
+        # Drop duplicates, MetaSWAP does not support duplicate connections
+        dataframe = dataframe.drop_duplicates(
+            subset=["mod_id", "svat", "layer"], keep="first"
+        )
```
Manangka, line 103: Should we give a warning that duplicate connections are found and that we are using the first one?

Manangka (approved):

imod/msw/coupler_mapping.py
JoerivanEngelen: Good thought. I think in this case users are not so interested in this, as they wouldn't necessarily expect duplicate entries in a coupling file in the first place.

PR #1463, opened 2025-02-26
Issue #1372 test regridder

### LINKED ISSUE #1372: Test if xugrid's and iMOD Python's Regridder behave similar enough
It would be good to test iMOD Python's regridder and xugrid's behave similar enough. 
If there are any noticeable differences these should be documented.

JoerivanEngelen:
Fixes #1372

# Description
See linked issue for all my testing, I think xugrid's regridders are a better replacement of ``imod.prepare.Regridder``! Time to report the public! There is one edge case which I documented in the changelog.

- Extends tests to also verify the old Regridder works the same as xugrids regridder. I didn't bother with a big refactor to add this as separate tests, as I foresee that we will remove these unittests altogether when the iMOD Python's Legacy Regridder class is removed.
- Extend documentation with additional pointers to xugrid's regridders.

# Checklist

- [x] Links to correct issue
- [ ] Unit tests were added


docs/api/changelog.rst
```diff
@@ -9,6 +9,16 @@ The format is based on `Keep a Changelog`_, and this project adheres to
 [Unreleased]
 ------------
 
+From this release on, we recommend using `xugrids regridding utilities
+<https://deltares.github.io/xugrid/examples/regridder_overview.html>`_ for
+regridding individual grids instead of :class:`imod.prepare.Regridder`. Xugrid's
+regridders are tested to be about 10 times faster than
+:class:`imod.prepare.Regridder`. There is one small difference: xugrid's
+``xugrid.BaryCentricInterpolator`` considers points cell edges to be inside,
```
Manangka, line 17: Could you clarify this a bit more. What points? Inside what?

imod/prepare/regrid.py
```diff
@@ -313,6 +315,22 @@ class Regridder(object):
     Object to repeatedly regrid similar objects. Compiles once on first call,
     can then be repeatedly called without JIT compilation overhead.
 
+    .. caution::
+
+        ``imod.prepare.Regridder`` is deprecated so we will remove it
```
Manangka, line 320: You can use the @depcrecated decorator to add a deprecation warning:
```
@deprecated("Use the regridder in ``xugrid`` instead")
class Regridder(object):
...

https://docs.python.org/3.13/library/warnings.html#warnings.deprecated
```

Manangka (approved):

docs/api/changelog.rst
JoerivanEngelen: Clarified to: 

_"There is one small difference: xugrid's
``xugrid.BaryCentricInterpolator`` considers sample points of the destination
grid that lie on the source grid's cell edges to be inside, whereas
:class:`imod.prepare.Regridder` considers them to be outside."_

imod/prepare/regrid.py
JoerivanEngelen: I tried this as it is a good idea, but for some reason it threw an error on generating the sphinx docs:

```
(exception: no module named imod.evaluate)
```

But normally sphinx has no issues with this, so I don't now what is happening here.

JoerivanEngelen: Created an issue for this, will merge this PR, is probably going to require more work https://github.com/Deltares/imod-python/issues/1468

docs/api/changelog.rst
Huite: Not entirely accurate, since the imod regridder only considers the right cell edge as outside; the left edge is inside.

This creates no problems with interior holes, since for two neighboring cells, the point will be assigned the value of the left cell. This only works for axis-aligned cell boundaries (since we can rely on direct comparison with `>=` or `<`), but doesn't work for non the axis-aligned cell boundaries that xugrid supports.

Anyway, close enough for me.

PR #1464, opened 2025-02-27
Issue #1462 move clip utilities

### LINKED ISSUE #1462: Move ``clip`` from ``imod.mf6.utilities`` to ``imod.common.utilities``
This issue blocks https://github.com/Deltares/imod-python/issues/1414

JoerivanEngelen:
Fixes #1462

# Description
- Moves ``clip`` utilities from ``imod.mf6.utilities`` to ``imod.common.utilities``
- Moves ``hfb`` utilities from ``imod.mf6.utilities`` to ``imod.common.utilities``
- Moves ``grid`` utilities from ``imod.mf6.utilities`` to ``imod.common.utilities``
- Move temporary removal and setting again of ``auxiliary variables`` (a MODFLOW6 affair) to the ``imod.mf6.Package.regrid_like`` method.
- Fix Sonarcloud issue: ``np.where(active > 0) == np.nonzero(active > 0)``, the latter is more explicit.

# Checklist
- [x] Links to correct issue
- [ ] Unit tests were added


imod/common/utilities/hfb.py
Manangka, line 1: Isn't this a mf6 specific utility? Or is it also needed for msw? or something else?
If it is mf6 only then i think you can better leave it in the mf6.utilitlies module

imod/mf6/package.py
```diff
@@ -624,7 +626,11 @@ def regrid_like(
         similar to the one used in input argument "target_grid"
         """
         try:
+            if hasattr(self, "auxiliary_data_fields"):
```
Manangka, line 629: Hmm this is also something that can be cleaned up using interfaces. In the future we ma y introduce an IAuxiliaryData interface

Manangka (approved):

imod/mf6/package.py
JoerivanEngelen: Might be a possiblity, I'm not sure as it would be a very MODFLOW6 specific interface. 

Other idea, somewhat relevant to your remark: What would reduce code complexity somewhat and be nice here is creating a decorator which removes auxiliary fields before calling a function and then adds them again. Still a bit on the fence, as I find decorators harder to debug.

imod/common/utilities/hfb.py
JoerivanEngelen: Done, renamed to ``imod/common/utilities/line_data.py`` also renamed some of the methods to have less confusing names.

PR #1485, opened 2025-03-12
Issue #356 Add test for a model with spaces in path

### LINKED ISSUE #356: MF6: Lacking support for spaces in path of model directory
In GitLab by @JoerivanEngelen on May 8, 2023, 11:26

A student ran into the issue where she ran her model in a model directory with a space in the path (default by OneDrive) 
This worked fine on iMOD Python 0.11.6. However, when we updated iMOD Python to 0.13.1, this did not work anymore.
No other packages got updated when we ran the ``mamba update`` command, except ``xugrid``, but this cannot have had any influence.

In general it is wise to not use spaces in paths to avoid trouble as a user; but there shouldn't be a reason for iMOD Python not to support spaces in paths.

JoerivanEngelen:
Fixes #356

# Description
- Refactor test for multiple directory cases
- Include case for directory with spaces

# Checklist
- [x] Links to correct issue
- [x] Unit tests were added

### DIFF
--- a/imod/tests/test_mf6/test_mf6_simulation.py
+++ b/imod/tests/test_mf6/test_mf6_simulation.py
@@ -183,6 +183,31 @@ def test_twri_disv_mdal_compliant_semi_roundtrip(twri_disv_model, tmpdir_factory
     assert len(ds.data_vars) == 3
 
 
+def test_simulation_open_head(circle_model, tmp_path):
+    simulation = circle_model
+
+    # Should throw error when model not run yet.
+    with pytest.raises(RuntimeError):
+        simulation.open_head()
+
+    modeldir = tmp_path / "circle"
+    simulation.write(modeldir)
+    simulation.run()
+    head = simulation.open_head()
+
+    assert isinstance(head, xu.UgridDataArray)
+    assert head.dims == ("time", "layer", "mesh2d_nFaces")
+    assert head.shape == (52, 2, 216)
+
+    # open heads with time conversion.
+    head = simulation.open_head(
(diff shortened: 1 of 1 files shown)


imod/tests/test_mf6/test_mf6_simulation.py
```diff
@@ -182,50 +183,43 @@ def test_twri_disv_mdal_compliant_semi_roundtrip(twri_disv_model, tmpdir_factory
...
-    assert head.dims == ("time", "layer", "mesh2d_nFaces")
-    assert head.shape == (52, 2, 216)
-    assert str(head.coords["time"].values[()][0]) == "2013-04-29T22:00:00.000000000"
+    def case_space_path(self, tmp_path):
+        return tmp_path / "dir with spaces"
 
 
-def test_simulation_open_head_relative_path(circle_model, tmp_path):
+@parametrize_with_cases("path", cases=PathCases)
+def test_simulation_open_head(circle_model, tmp_path, path):
```
Manangka, line 198: I would create a separate test for testing the path cases. You wouldn't need to open the heads, just testing if it can write and run without an error would be enough

It also makes it clear what a test tests. Now it is testing 2 things (opening heads and testing different path formats)

Manangka (approved):

imod/tests/test_mf6/test_mf6_simulation.py
JoerivanEngelen: Agreed. I separated this into two tests, simplifying them. I kept the call to ``open_head`` in the tests for different paths though, as ``open_head`` also requires some logic in constructing paths to the model output files (in this case the .hds and .grb files).

PR #1493, opened 2025-03-19
Issue #1492 migrate from fastcore to plum

### LINKED ISSUE #1492: Error upon import with latest fastcore
I saw all PipPython jobs failing in the master branch, this is because fastcore just had an update. Apparently we need to migrate our typedispatching to ``plum``

https://answerdotai.github.io/fasttransform/fastcore_migration_guide.html

Shouldn't be a lot of work.

JoerivanEngelen:
Fixes #1492

# Description
Replace fastcore typedispatching with plum's dispatching.
Plum is a bit stricter, which I generally think is good. When fastcore's typedispatching didn't encounter a typedispatched fallback function for type ``Any``, for some reason things worked, but I'm not sure what it did. Plum is stricter so I had to explicitly had to add some fallback functions to either raise an error or do something.

In detail, this PR
- Replaces @typedispatch with a plum Dispatcher instance, this also limits the scope for accidentily adding typedispatched functions in other modules
- Separate ``clip_by_grid`` method for GridDataArray into one for xr.DataArray and xu.UgridDataArray, as plum apparently thinks this ambiguous.
- Update some arrays in ``create_exchange_package`` to become DataArray
- Add missing fallback functions for ``enforce_dim_order`` and ``is_unstructured``
- Add function for ``enforce_dim_order`` with ``None``
- Change fallback type from ``object`` to ``Any``, to also support ``None``

# Checklist

- [x] Links to correct issue
- [ ] Unit tests were added


imod/common/utilities/clip.py
```diff
@@ -88,6 +91,33 @@ def clip_by_grid(  # noqa: F811
     return cls._from_dataset(selection)
 
 
+@dispatch  # type: ignore[no-redef, misc]
+def clip_by_grid(package: ILineDataPackage, active: xr.DataArray) -> ILineDataPackage:  # noqa: F811
+    """Clip LineDataPackage outside structured grid."""
+    return _clip_by_grid_line_data(package, active)
+
+
+# For some reason the plum dispatching finds "active: GridDataArray" ambiguous
```
Manangka, line 100: Maybe raise an issue about this on the their board? Seems like something that should be supported

Manangka (approved):

imod/common/utilities/clip.py
JoerivanEngelen: I don't have time now to further investigate this. Seems somewhat related to this https://github.com/beartype/plum/issues/149.

PR #1497, opened 2025-03-26
Stream Flow Routing (SFR)

Huite:
Still work in progress, but adds most of the required logic for an SFR package.

This requires an unreleased xugrid Ugrid1d method, but otherwise it looks like the Ugrid1d class suffices.

One of the main things that requires a decision is how to assign the network to layers. Unlike a river or drainage boundary, it doesn't seem like a great idea idea to assign a single water body to multiple cells (even if intersects multiple layers), because the SFR package computes a single water height per reach.

I think the easiest solution is to add a layer variable. If this isn't provided (`None`), I guess we'd use the streambed elevation to assign to a layer instead in combination with DIS top and bottom data.

It would also be good to have another critical look at the AdvancedBoundary class. It looks like they have very little in common, so we should probably just make that an abstract base class which requires the `write_blockfile` method, but not much else. Also worth researching is how where external files are allowed (which apparently works for UZF?). This just dumps to the blockfile, like the lake package implementation.

The melting approach for the period data works to some satisfaction, it looks better to me than the current period jinja stuff in the lake package.

### DIFF
--- a/imod/mf6/disv.py
+++ b/imod/mf6/disv.py
@@ -131,18 +131,19 @@ class VerticesDiscretization(Package, IRegridPackage, IMaskingSettings):
         return df
 
     def _cell2d_dataframe(self) -> pd.DataFrame:
+        XUGRID_FILL = -1
         grid = self.dataset.ugrid.grid
         df = pd.DataFrame(grid.face_coordinates)
         df.index += 1
         # modflow requires clockwise; ugrid requires ccw
         face_nodes = grid.face_node_connectivity[:, ::-1]
-        df[2] = (face_nodes != grid.fill_value).sum(axis=1)
+        df[2] = (face_nodes != XUGRID_FILL).sum(axis=1)
         for i, column in enumerate(face_nodes.T):
             # Use extension array to write empty values
             # Should be more efficient than mixed column?
             df[3 + i] = pd.arrays.IntegerArray(
                 values=column + 1,
-                mask=(column == grid.fill_value),
+                mask=(column == XUGRID_FILL),
             )
         return df
 
--- a/imod/mf6/sfr.py
+++ b/imod/mf6/sfr.py
@@ -1,0 +1,657 @@
+from typing import Literal, Union
+
+import numpy as np
+import pandas as pd
+import xarray as xr
+
+from imod.logging import init_log_decorator
+from imod.mf6.boundary_condition import AdvancedBoundaryCondition, BoundaryCondition
+from imod.mf6.write_context import WriteContext
+from imod.schemata import (
+    AllValueSchema,
+    DimsSchema,
+    DTypeSchema,
+)
+from imod.select.points import points_indices
+from imod.typing.grid import GridDataArray
+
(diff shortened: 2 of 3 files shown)


Huite:
I've created a basic example, which seems to give reasonable results (I've only looked at heads, though).

The most controversial thing so far is that just like the HFB and Well classes, this is a grid agnostic package. So it needs information from DIS (a dst_grid) to place the elements. Placing the elements can be done in a simple way, by simply placing them where the edge centroid is located (alternatively grid cell with largest overlap). (Note that an SFR reach can only be connected to one cell at most, subdivision should be done via regridding.) Anyway, I've been lazy so far, and I just added a cellid variable to the existing class so I don't have to duplicate everything.
Having a separate class is probably best, because it's what you'd want to represent e.g. bare input as well (without the network topology).

Some tedious things:

* https://github.com/Deltares/imod-python/issues/1498
* What's the best way to check for `None` for optional values? In the lake package there are some poor checks still, by the way (`da.values[()] is None`).

Huite:
I've added some basic functions to read the output stages and budgets as well.

There is clear opportunity here to refactor the output reading stuff. What makes things a bit more complicated:

* We have data that makes sense on the model grid. This is true for most packages, like NPF flows, but also in how we treat rivers, drainage, etc.
* This is also true for the UZF budget logic that @HendrikKok added/had to make slight modifications for.
* In principle, the SFR data is valuable on both the grid (summed values for reaches), but also on the original network topology.
* A similar argument applies to multi-aquifer-wells (MAW) which you'd probably like to get as point data too.

I originally chose some dict based dispatch:

```python
_READ_GRB = {
    "grid dis": dis.read_grb,
    "grid disv": disv.read_grb,
    "grid disu": disu.read_grb,
}

_OPEN_HDS = {
    "dis": dis.open_hds,
    "disv": disv.open_hds,
    "disu": disu.open_hds,
}

_OPEN_CBC: Dict[str, Callable] = {
    "dis": dis.open_cbc,
    "disv": disv.open_cbc,
    "disu": disu.open_cbc,
}

_OPEN_DVS = {
    "dis": dis.open_dvs,
    "disv": disv.open_dvs,
    "disu": disu.open_dvs,
}
```

I think this should be simply replaced by objects and inheritance, letting Python do the method dispatch in the regular manner.

We could then have separate readers; one for DIS, DISV, DISU, SFR, etc.

JoerivanEngelen:
> Some tedious things:
> 
> * [mf6 Package: copy method is broken #1498](https://github.com/Deltares/imod-python/issues/1498)
> * What's the best way to check for `None` for optional values? In the lake package there are some poor checks still, by the way (`da.values[()] is None`).
> 

Regarding these things:

* The copy method has now been fixed
* For your latter question, we unfortunately don't have a function you could direct use now. But we can draw inspiration from the ``_is_scalar_nan`` helper function, I reckon. https://github.com/Deltares/imod-python/blob/c1adb9eb414b25ec3472a8b5c9b949870976c0fb/imod/mf6/pkgbase.py#L24

PR #1505, opened 2025-04-16
Issue #1498 mf6 Package: copy method is broken

### LINKED ISSUE #1498: mf6 Package: copy method is broken
```python
nlayer = 1
nrow = 5
ncol = 5

dx = 1000.0
dy = -1000.0

x = np.arange(0, dx * ncol, dx) + 0.5 * dx
y = np.arange(0, dy * nrow, dy) + 0.5 * dy
layer = np.array([1])

idomain = xr.DataArray(np.ones((nlayer, nrow, ncol), dtype=int), dims=("layer", "y", "x"), coords={"layer": layer, "y": y, "x": x})
rch_rate = xr.full_like(idomain.sel(layer=1), 0.001, dtype=float)

rch = imod.mf6.Recharge(rch_rate)
rch.copy()
```

Results in:
```python
TypeError: Recharge.__init__() got an unexpected keyword argument 'coords'
```

Problem is:

```python
    def copy(self) -> Any:
        # All state should be contained in the dataset.
        return type(self)(**self.dataset.copy().to_dict())
```

to_dict() will insert a coords member, which obviously won't be a keyword for a package.

verkaik:
Fixes #1498 

# Description

PR for @verkaik to push his bug fix and additional unittest(s) to support this.
# Checklist

- [x] Links to correct issue
- [x] Unit tests were added

### DIFF
--- a/docs/api/changelog.rst
+++ b/docs/api/changelog.rst
@@ -23,6 +23,8 @@ Fixed
   :meth:`imod.mf6.GeneralHeadBoundary.clip_box` threw an error when
   ``time_start`` or ``time_end`` were set to ``None`` and a ``"repeat_stress"``
   was included in the dataset.
+- Fixed bug where :meth:`imod.mf6.package.copy` threw an error.
+
 
 [1.0.0rc2] - 2025-03-05
 -----------------------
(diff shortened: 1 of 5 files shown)


JoerivanEngelen (approved):
Thanks,

I took the liberty of fixing the mypy issue and adding some unittests, for both structured as well as unstructured models.

PR #1514, opened 2025-04-22
Add an argument to quote paths for iMOD-WQ

Huite:
For Teun

### DIFF
--- a/imod/wq/bas.py
+++ b/imod/wq/bas.py
@@ -82,18 +82,20 @@ class BasicFlow(Package):
                 f" got instead {dims}"
             )
 
-    def _render(self, directory, nlayer, *args, **kwargs):
+    def _render(self, directory, nlayer, quote=False, *args, **kwargs):
         """
         Renders part of runfile that ends up under [bas] section.
         """
         d = {}
         for varname in ("ibound", "starting_head"):
-            d[varname] = self._compose_values_layer(varname, directory, nlayer)
+            d[varname] = self._compose_values_layer(
+                varname, directory, nlayer, quote=quote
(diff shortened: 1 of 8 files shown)


JoerivanEngelen:
Thanks for trying to hotfix this. FYI: This is not an immediate problem anymore: I just talked to @TeunvanWoerkom, he moved to a drive without spaces in the path.

I'm not sure if we should merge this, as the ``wq`` test bench doesn't have the coverage of the ``mf6`` coverage. Most notably, we don't run a model with iMOD-WQ to test if it breaks. The current ``wq`` module is tried and tested in many projects, so that is sufficient to do work with. I don't want to risk breaking it with changes.

Besides, he manually quoted the paths in his runfile to fix this and then iMOD-WQ apparently also isn't able to deal with spaces in paths (at least the version he used). It crashes upon attempting to create the results directories.

Alternatively: We could raise an error if a user attempts to write a ``wq`` model in a folder with a space in the path?

Huite:
> Besides, Teun manually quoted the paths in his runfile to fix this and then iMOD-WQ apparently also isn't able to deal with spaces in paths (at least the version he used). It crashes upon attempting to create the results directories.

Hah! That changes things. It means that the somewhat nice error he reported is probably only partially true; i.e some part of the program is space aware, but not everything.

Emitting a warning or raising an error is much simpler. Adding this new argument was a tedious affair (luckily it's packaged in the MODFLOW 6 logic WriteContext).

JoerivanEngelen:
Followed up in: https://github.com/Deltares/imod-python/pull/1517

PR #1517, opened 2025-04-28
Issue #305 Throw error upon space in directory name.

### LINKED ISSUE #305: WQ: Spaces in paths in runfile: quotes should be added
In GitLab by @gmcmjanssen on Jan 4, 2023, 18:04

I noticed that when a model is put in a folder that has spaces in its name (e.g. in the Deltares OneDrive folder), iMOD-Python does not put the paths between quotes. This then results in an error in iMOD-WQ.

E.g:
[bas6]
    ibound_l$ = c:/Users/janssen_gs/OneDrive - Stichting Deltares/DATA/PROMISCES/Henry/data/3-input/bas/ibound_l$.idf

Should be 

[bas6]
    ibound_l$ = "c:/Users/janssen_gs/OneDrive - Stichting Deltares/DATA/PROMISCES/Henry/data/3-input/bas/ibound_l$.idf"

JoerivanEngelen:
Closes #305 

# Description

This throws an error when a user tries to write a model in a directory with a space in it.

Follow-up to https://github.com/Deltares/imod-python/pull/1514

See my comment here: https://github.com/Deltares/imod-python/pull/1514#issuecomment-2829863678

> Thanks for hotfixing this. FYI: This is not an immediate problem anymore: I just talked to @TeunvanWoerkom, he moved to a drive without spaces in the path.
> 
> I'm not sure if we should merge this, as the `wq` test bench doesn't have the coverage of the `mf6` module. Most notably, we don't run a model with iMOD-WQ to test if it breaks. The current `wq` module is tried and tested in many projects, so that is sufficient to do work with. I don't want to risk breaking it with changes.
> 
> Besides, Teun manually quoted the paths in his runfile to fix this and then iMOD-WQ apparently also isn't able to deal with spaces in paths (at least the version he used). It crashes upon attempting to create the results directories.
> 
> Alternatively: We could raise an error if a user attempts to write a `wq` model in a folder with a space in the path?

It turned out that adding quotes doesn't do the trick, as iMOD-WQ also cannot handle spaces in paths, even when they are quoted in the runfile. It crashes upon writing results.

# Checklist

- [x] Links to correct issue
- [x] Unit tests were added

### DIFF
--- a/docs/api/changelog.rst
+++ b/docs/api/changelog.rst
@@ -6,6 +6,15 @@ All notable changes to this project will be documented in this file.
 The format is based on `Keep a Changelog`_, and this project adheres to
 `Semantic Versioning`_.
 
+[Unreleased]
+------------
+
+Changed
+~~~~~~~
+
+- :meth:`imod.wq.SeawatModel.write` now throws an error if trying to write in a
+  directory with a space in the path. (iMOD-WQ does not support this.)
+
 [1.0.0rc3] - 2025-04-17
 -----------------------
 
--- a/imod/tests/test_wq/test_wq_model.py
+++ b/imod/tests/test_wq/test_wq_model.py
@@ -772,3 +772,26 @@ def test_write_result_dir_is_workdir(basicmodel, tmp_path):
 
     assert line.split("=")[-1].strip() == "."
     # TODO: more rigorous testing
+
+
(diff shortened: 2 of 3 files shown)


Huite (approved):
lgtm

PR #1518, opened 2025-04-30
Issue #1515 fix xorigin yorigin disv

### LINKED ISSUE #1515: [Bug] - wrong xorigin,yorigin in DISV
It appears that an error is made when writing discretization files (dis,disv). The parameters XORIGIN and YORIGIN are user configurable offsets to the coordinates used to set up the grid. Currently they are written as the minimum value of x and y values, eg for DISV:

```
d["xorigin"] = grid.node_x.min()
d["yorigin"] = grid.node_y.min()
```

**Steps to reproduce**
Write any discretization from imod-python to file, I'd_ expect.

JoerivanEngelen:
Fixes #1515

# Description
Adds the following:

- Hardcode xorigins and yorigins in DISV package to 0.0.
- Add tests to verify if the xorigins and yorigins and DIS and DISV packages are rendered correctly
- Offset the grid in the DIS and DISV unittests, to verify if xorigins and yorigins are properly rendered.
- Account for xorigin and yorigin in the disv grb read funtion
- Add a test which has an offset in DISV, to see if our read DISV grb function actually properly deals with that. This is in the case somebody wants to read model output of a DISV model generated with Flopy that has an offset in xorigins and yorigins.
- Small refactor to override ``_get_render_dictionary`` instead of ``render`` in DIS and DISV
- Small refactor, move append of vertices and cell2d data to blockfile to a separate method.

# Checklist

- [x] Links to correct issue
- [x] Unit tests were added

### DIFF
--- a/imod/mf6/disv.py
+++ b/imod/mf6/disv.py
@@ -106,8 +106,8 @@ class VerticesDiscretization(Package, IRegridPackage, IMaskingSettings):
         disdirectory = directory / pkgname
         d = {}
         grid = self.dataset.ugrid.grid
-        d["xorigin"] = grid.node_x.min()
-        d["yorigin"] = grid.node_y.min()
+        d["xorigin"] = 0.0
+        d["yorigin"] = 0.0
         d["nlay"] = self.dataset["idomain"].coords["layer"].size
         facedim = grid.face_dimension
         d["ncpl"] = self.dataset["idomain"].coords[facedim].size
--- a/imod/mf6/disv.py
+++ b/imod/mf6/disv.py
@@ -102,7 +102,7 @@ class VerticesDiscretization(Package, IRegridPackage, IMaskingSettings):
         super().__init__(dict_dataset)
         self._validate_init_schemata(validate)
 
-    def render(self, directory, pkgname, globaltimes, binary):
+    def _get_render_dictionary(self, directory, pkgname, globaltimes, binary):
         disdirectory = directory / pkgname
         d: dict[str, Any] = {}
         grid = self.dataset.ugrid.grid
@@ -122,7 +122,7 @@ class VerticesDiscretization(Package, IRegridPackage, IMaskingSettings):
         d["idomain_layered"], d["idomain"] = self._compose_values(
             self["idomain"], disdirectory, "idomain", binary=binary
         )
-        return self._template.render(d)
+        return d
 
     def _verts_dataframe(self) -> pd.DataFrame:
         grid = self.dataset.ugrid.grid
--- a/imod/mf6/disv.py
+++ b/imod/mf6/disv.py
@@ -1,4 +1,4 @@
-from typing import List
+from typing import List, Any
 
 import numpy as np
 import pandas as pd
@@ -104,7 +104,7 @@ class VerticesDiscretization(Package, IRegridPackage, IMaskingSettings):
 
     def render(self, directory, pkgname, globaltimes, binary):
         disdirectory = directory / pkgname
-        d = {}
+        d: dict[str, Any] = {}
         grid = self.dataset.ugrid.grid
         d["xorigin"] = 0.0
         d["yorigin"] = 0.0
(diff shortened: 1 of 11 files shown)


Huite:
I would strongly suggest to also fix this:

https://github.com/Deltares/imod-python/blob/afcf5412b7941ce62dc02d64e8c42c86752c570e/imod/mf6/out/disv.py#L56
https://github.com/Deltares/imod-python/blob/afcf5412b7941ce62dc02d64e8c42c86752c570e/imod/mf6/out/disv.py#L57

Then add the values to the vertices:

https://github.com/Deltares/imod-python/blob/afcf5412b7941ce62dc02d64e8c42c86752c570e/imod/mf6/out/disv.py#L63


Note that we're skipping the centroid coordinates as well:

https://github.com/Deltares/imod-python/blob/afcf5412b7941ce62dc02d64e8c42c86752c570e/imod/mf6/out/disv.py#L64
https://github.com/Deltares/imod-python/blob/afcf5412b7941ce62dc02d64e8c42c86752c570e/imod/mf6/out/disv.py#L65

This should be the least controversial since a centroid is a centroid (for a convex polygon), but as far as I remember, xugrid won't respect these coordinates anyway:
https://github.com/Deltares/xugrid/issues/56

imod/mf6/disv.py
```diff
@@ -104,10 +104,10 @@ def __init__(self, top, bottom, idomain, validate: bool = True):
 
     def render(self, directory, pkgname, globaltimes, binary):
         disdirectory = directory / pkgname
-        d = {}
+        d: dict[str, Any] = {}
         grid = self.dataset.ugrid.grid
-        d["xorigin"] = grid.node_x.min()
-        d["yorigin"] = grid.node_y.min()
+        d["xorigin"] = 0.0
+        d["yorigin"] = 0.0
```
Manangka, line 110: I'm a bit confused. When partitioning a model, the partioned models should have same the same xorigin and yorigin as the original model right? I don't see where that is being done

JoerivanEngelen: The DISV package requires explicit coordinates of where the vertices are located in the VERTICES block in its MODFLOW6 input file. We supply the UgridDataArray vertices here, which contain the full coordinates, so xorigin and yorigin need to be 0.0, as the DISV package contains a direct representation of our grid coordinates.

The DIS package, however, doesn't have such a block, only "DELR" and "DELC" data in the GRIDDATA block. We therefore rely on xorigin an yorigin here to do the bookkeeping of the coordinates.

JoerivanEngelen:
> I would strongly suggest to also fix this while you're at it:
> 
> https://github.com/Deltares/imod-python/blob/afcf5412b7941ce62dc02d64e8c42c86752c570e/imod/mf6/out/disv.py#L56
> 
> 
> https://github.com/Deltares/imod-python/blob/afcf5412b7941ce62dc02d64e8c42c86752c570e/imod/mf6/out/disv.py#L57
> 
> Then add the values to the vertices:
> 
> https://github.com/Deltares/imod-python/blob/afcf5412b7941ce62dc02d64e8c42c86752c570e/imod/mf6/out/disv.py#L63
> 
> The current logic will work alright when xorigin and yorigin are 0 (as imod-python will do), but it fail when somebody has written their model e.g. with flopy with a nonzero origin.
> 
> Note that we're skipping the centroid coordinates as well:
> 
> https://github.com/Deltares/imod-python/blob/afcf5412b7941ce62dc02d64e8c42c86752c570e/imod/mf6/out/disv.py#L64
> 
> 
> https://github.com/Deltares/imod-python/blob/afcf5412b7941ce62dc02d64e8c42c86752c570e/imod/mf6/out/disv.py#L65
> 
> This should be the least controversial since a centroid is a centroid (for a convex polygon), but as far as I remember, xugrid won't respect these coordinates anyway: [Deltares/xugrid#56](https://github.com/Deltares/xugrid/issues/56)

Good point, I've added this and added a test which overrides a DISV file of a simulation, runs it, reads the CBC and checks if the offset in the origin is accounted for.

Manangka (approved):

mjr-deltares:
I missed the review request for this issue, sorry 'bout that. (But all good as far as I can tell)

Issue #668, opened 2023-11-24
implement metis-based partitioning for structured grids


Manangka:
In GitLab by @luitjansl on Nov 24, 2023, 10:24

We currently use metis for unstructured grids, but we would like to also use it for structured grids. 

Some pointers on how to do this can be found below

![image.png](https://gitlab.com/deltares/imod/imod-python/uploads/056f8d570820ac148aced994d58bc192/image.png)

PR #1520, opened 2025-05-01
Issue #1519 weights in label arrays

### LINKED ISSUE #1519: [FEATURE] - Allow users to set weights when creating label arrays for partitioning
``xugrid`` 0.14.0 just released and it includes an argument to provide weights to METIS for partitioning.

I think we can let the user provide custom weights as an argument, to balance computation loads better. The following code is sufficient for most cases:

```python
weights = (idomain == 1).sum("layer").astype(int)
```

I think this can be the default if no weights are provided by the user. 

This is trivial to implement for DISV (for which we have a project now running), but harder to implement for DIS, as it  requires overhauling the now rudimentary DIS partition generator. Probably having something akin to what was in iMOD-WQ with a recursive coordinate bisection algorithm is sufficient.

JoerivanEngelen:
Fixes #1519 and #668

# Description
Implements the following:

- Provide weights to xugrid -> PyMetis calls
- Also call PyMetis for structured grids, instead of the naive uniform splitter without options for weighting grids.
- Change name of label array to "label" instead of "idomain"
- Move ``get_label_array`` to ``imod.prepare.create_partition_labels``, and add this to public API. Calling the old ``get_label_array`` results in a DeprecationWarning, after which args are passed on to  ``imod.prepare.create_partition_labels``.
- Fix bug where arrays provided to exchange creator do not have to be named "idomain" anymore. They are now renamed to enforce a name.
- Pin xarray to 2025.3.0 to avoid error and HFB package with \_\_repr\_\_

I was still on the fence about the latter, on one hand, I like that these changes as they keep the code paths very similar and support providing weights, on the other hand, the way Metis is now called, it results in sloppy looking label arrays for structured grids.

For this [trivial example](https://deltares.github.io/imod-python/examples/mf6/transport_2d.html) with idomain equal to one everywhere. We had this:

image

and with METIS we get this:

image

That is still somewhat acceptable, though confusing as it looks sloppy, so it deserves some extra explanation.
Perhaps manually create a simple label array here is better?

For the hondsrug example, which is not a perfect square model, METIS accounts for the fact that the model domain is rectangular, which the previous implementation does not: 

image

So in the end I think it is an improvement

# Checklist

- [x] Links to correct issue
- [x] Unit tests were added

### DIFF
--- a/imod/prepare/partition.py
+++ b/imod/prepare/partition.py
@@ -5,7 +5,7 @@ import xarray as xr
 import xugrid as xu
 from plum import Dispatcher
 
-from imod.mf6.simulation import Modflow6Simulation
+from imod.common.interfaces.isimulation import ISimulation
 from imod.typing import GridDataArray
 from imod.typing.grid import as_ugrid_dataarray
 
@@ -53,7 +53,7 @@ def _partition_idomain(weights: xr.DataArray, npartitions: int) -> GridDataArray
 
 
 def create_partition_labels(
-    simulation: Modflow6Simulation,
+    simulation: ISimulation,
     npartitions: int,
     weights: Optional[GridDataArray] = None,
 ) -> GridDataArray:
--- a/imod/prepare/partition.py
+++ b/imod/prepare/partition.py
@@ -1,0 +1,95 @@
+import copy
+from typing import Optional
+
+import xarray as xr
+import xugrid as xu
+from plum import Dispatcher
+
+from imod.mf6.simulation import Modflow6Simulation
+from imod.typing import GridDataArray
+from imod.typing.grid import as_ugrid_dataarray
+
+# create dispatcher instance to limit scope of typedispatching
+dispatch = Dispatcher()
+
+
+def _validate_weights(idomain: GridDataArray, weights: GridDataArray) -> None:
+    idomain_top = copy.deepcopy(idomain.isel(layer=0))
+    if not isinstance(weights, type(idomain)):
+        raise TypeError(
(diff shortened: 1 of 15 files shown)


imod/prepare/partition.py
```diff
@@ -0,0 +1,95 @@
...
+
+@dispatch  # type: ignore[no-redef]
+def _partition_idomain(weights: xr.DataArray, npartitions: int) -> GridDataArray:
+    """
+    Convert to UgridDataArray to use xugrid to call PyMetis to create a label
+    array for structured grids. Call as_ugrid_dataarray to used cached results
+    of ``Ugrid2d.from_structured`` if available, to save some costly
+    conversions.
+    """
+    weights_uda = as_ugrid_dataarray(weights)
```
Manangka, line 46: I'm wondering what the performance impact is of converting this. Have you tested it? Models that are being split are usually big

Manangka (approved):

imod/prepare/partition.py
JoerivanEngelen: Good point. I've tested this before, as we also convert structured grids to unstructured grids it for the horizontal flow barrier package (to snap to edges). There, it posed a performance hit as the conversion had to be called multiple times (the LHM has ~20 HFB packages). That's why ``as_ugrid_dataarray`` caches the result of this conversion, so it doesn't have to be called multiple times. I think for the LHM (a very large model > 20M cells), it is bearable to call the conversion once. It takes about a minute for that.

PR #1533, opened 2025-05-20
Render inactive cells with rates as "" instead of 0.0

### LINKED ISSUE #1532: [Bug] - Incorrect writing of MetaSWAP file with sprinkling capacities (scap_svat.inp)
**Bug description**
- MetaSWAP svats with surface water sprinkling activated do not seem to be written to scap_svat.inp
- default values are written, e.g. as 0., but this likely to give issues with the MetaSWAP kernel that checks fixed-sized strings for not being empty to increase internal counters (not yet tested). 

**Steps to reproduce**
Use a LHM iMOD project (.prj) file with irrigation data specified in the (CAP) section.

JoerivanEngelen:
Fixes #1532

# Description
Render inactive cells in ``scap_svat.inp`` as "" instead of 0.0. I'm not sure if this is entirely necessary as it complicates the nice fixed format writers.

# Checklist
- [x] Links to correct issue
- [x] Unit tests were added

### DIFF
--- a/imod/msw/fixed_format.py
+++ b/imod/msw/fixed_format.py
@@ -15,11 +15,15 @@ class VariableMetaData:
 
 
 def format_fixed_width(value, metadata):
-    if metadata.dtype == str:
+    if value == "":
+        col_width = metadata.column_width - 2
+        format_string = "{:" + f"{col_width}" + '}""'
+        return format_string.format(value)
+    if metadata.dtype is str:
         format_string = "{:" + f"{metadata.column_width}" + "}"
-    elif metadata.dtype == int:
+    elif metadata.dtype is int:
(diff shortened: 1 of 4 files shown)


Huite:
Note that pandas has some `NA` support.

E.g. for the MODFLOW 6 input files which have an irregular number of columns (e.g. DISV), I use the pandas integer NA. https://pandas.pydata.org/docs/user_guide/integer_na.html

You might be able to use some comparable functionality.

JoerivanEngelen:
I refrained from adding the extra logic to "" marks instead of 0.0 in scap_svat, as @HendrikKok and I don't see any reason why MetaSWAP would fail when a 0.0 is written in ``scap_svat.inp`` instead of "".

PR #1562, opened 2025-06-13
Issue #1561 improve performance splitting

### LINKED ISSUE #1561: Poor performance splitting simulations
I've debugged @TeunvanWoerkom's model further and found something really slowing down things: 

For structured grids and HFB packages, the splitting can result into empty packages, so there is a check to ``is_empty``. This for each variable checks there is any active data. There is some messing around with inverts going on, so this could cause things to slow down. 

I think we can do the following things to speed things up:

- Only call is_empty if HFB or package on a structured grid
- Investigate if the AllNoDataScheme which is used to check is_empty can be speeded up.

JoerivanEngelen:
Fixes #1561

# Description
- Add option ``ignore_time_purge_empty`` to only check ``imod.mf6.Modflow6Simulation.split`` the first timestep during the purge of empty packages. This saves a lot of time evaluating models with multiple stress periods.
- Fix bug where an error was thrown in :class:`imod.mf6.Well` when an entry had to be filtered and its ``id`` didn't match the index.
- Refactor: Call ``purge_empty_packages`` instead of separate logic for removing empty packages in ``imod.mf6.Modflow6Simulation.split``
- Performance: During splitting, only mask grids when structured, for unstructured grids this is unnecessary.
- Refactor: Remove unnecessary duplicate masking for structured grids when splitting

# Checklist

- [x] Links to correct issue
- [x] Unit tests were added

### DIFF
--- a/imod/mf6/multimodel/modelsplitter.py
+++ b/imod/mf6/multimodel/modelsplitter.py
@@ -118,5 +118,5 @@ def slice_model(partition_info: PartitionInfo, model: IModel) -> IModel:
             new_model[pkg_name] = sliced_package
 
         if isinstance(package, BoundaryCondition):
-            expand_transient_auxiliary_variables(package)
+            expand_transient_auxiliary_variables(sliced_package)
     return new_model
--- a/imod/mf6/multimodel/modelsplitter.py
+++ b/imod/mf6/multimodel/modelsplitter.py
@@ -64,22 +64,6 @@ def _validate_submodel_label_array(submodel_labels: GridDataArray) -> None:
         )
 
 
-def mask_if_structured(
-    package: Package,
-    active_domain: GridDataArray,
-) -> Optional[Package]:
-    """
-    Masks the package if it is structured. If the package is unstructured, it
-    returns the package unchanged.
-    """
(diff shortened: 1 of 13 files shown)


imod/mf6/multimodel/modelsplitter.py
```diff
@@ -96,17 +113,10 @@ def slice_model(partition_info: PartitionInfo, model: IModel) -> IModel:
             remove_expanded_auxiliary_variables_from_dataset(package)
 
         sliced_package = clip_by_grid(package, partition_info.active_domain)
-
-        sliced_package = sliced_package.mask(new_idomain)
-        # The masking can result in packages with AllNoData.Therefore we'll have
-        # to drop these packages.
-        if not sliced_package.is_empty():
+        maybe_masked_package = mask_if_structured(sliced_package, new_idomain)
+        if maybe_masked_package is not None:
```
Manangka, line 117: These 2 lines read a bit awkward. What exactly are you trying to test?
If it is structured and ... then assign new sliced package

What makes it more confusing is that in the mask_if_structured method you also have an action when the grid is unstructured

PR #1591, opened 2025-07-11
Issue #1580 add reallocate method

### LINKED ISSUE #1580: [FEATURE] - Add ``reallocate`` method
I noticed that when upscaling models, quite often topsystems need to be reallocated. See the example I just added https://github.com/Deltares/imod-python/pull/1579 

Right now there is no method to directly aggregate layered grids into planar grids and allocating and distributing conductances again.

The packages which can be reallocated should have some object containing aggregation methods for each variable. Most variables require a ``np.nanmean``, but the conductances require a ``np.nansum``.

I also noticed that for the drain we might need something extra, as iMOD5 river packages with an infiltration factor are split into river and drain packages. In this case, we would like to be able to reallocate the drain package in a similar manner as the river package, but we lack the bottom elevations for this. The ``Drainage.reallocate`` method should therefore accept an optional river bottom elevation grid to also support the River allocation and distributing options.

JoerivanEngelen:
Fixes #1580

# Description
This PR changes the following:

- Add ``reallocate`` method
- Update examples with new method
- Convert aggregate method to classmethod as this allows some flexibility for calling this method elsewhere.
- Rename ``_get_options`` to ``_get_pkg_options``
- Separate logic to get options into ``_get_unfiltered_pkg_options``
- Boyscouting: Type annotation return types in pkgbase
- Boyscouting: Add missing default option for rch in SimulationAllocationOptions

# Checklist
- [x] Links to correct issue
- [x] Unit tests were added

### DIFF
--- a/imod/tests/test_mf6/test_mf6_drn.py
+++ b/imod/tests/test_mf6/test_mf6_drn.py
@@ -244,7 +244,7 @@ def test_3d_singelayer():
 def test_aggregate_layers(drainage):
     river = imod.mf6.Drainage(**drainage)
 
-    planar_dict = river.aggregate_layers()
+    planar_dict = river.aggregate_layers(river.dataset)
     assert isinstance(planar_dict, dict)
     for value in planar_dict.values():
         assert isinstance(value, xr.DataArray)
--- a/imod/tests/test_mf6/test_mf6_rch.py
+++ b/imod/tests/test_mf6/test_mf6_rch.py
@@ -204,7 +204,7 @@ def test_transient_no_layer_dim(rch_dict_transient):
 
 def test_transient_aggregate(rch_dict_transient):
     rch = imod.mf6.Recharge(**rch_dict_transient)
-    planar_dict = rch.aggregate_layers()
+    planar_dict = rch.aggregate_layers(rch.dataset)
 
     assert isinstance(planar_dict, dict)
     for value in planar_dict.values():
--- a/imod/tests/test_mf6/test_mf6_riv.py
+++ b/imod/tests/test_mf6/test_mf6_riv.py
@@ -416,7 +416,7 @@ def test_aggregate_layers(riv_data):
         else xr.DataArray
     )
 
-    planar_dict = river.aggregate_layers()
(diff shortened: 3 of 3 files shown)


examples/user-guide/09-topsystem.py
```diff
@@ -434,5 +435,45 @@
 
 riv
 
+# %%
+#
+# Reallocate package
+# ------------------
+#
+# The river package has a :meth:`imod.mf6.River.reallocate` method which can be
```
Manangka, line 443: You explain that there is a reallocte method but it might also be good to explain why you are using it

imod/mf6/boundary_condition.py
```diff
@@ -298,19 +338,107 @@ def aggregate_layers(self) -> GridDataDict:
...
+            A new instance of the boundary condition class with the reallocated
+            data. The original instance remains unchanged.
+        """
+        allocation_option, distributing_option = _handle_reallocation_options(
+            self._pkg_id, allocation_option, distributing_option
+        )
+        # Aggregate data to planar data first
+        planar_data = self.aggregate_layers(self.dataset)
+        if "conductance" in self.dataset.data_vars:
+            if npf is None:
```
Manangka, line 398: You can move this check to the start of this method. That way you can fail fast and don't have to call aggregate_layers

```
...
 BoundaryCondition
            A new instance of the boundary condition class with the reallocated
            data. The original instance remains unchanged.
        """
if "conductance" in self.dataset.data_vars and npf is None:
  raise ValueError(
                    "NodePropertyFlow must be provided "
                    "for packages with conductance variable."
                )

 # Aggregate data to planar data first
  planar_data = self.aggregate_layers(self.dataset)

...
```

Manangka (approved):

imod/mf6/boundary_condition.py
JoerivanEngelen: Good point, thanks!

examples/user-guide/09-topsystem.py
JoerivanEngelen: Good point, added some extra explanation.

PR #1628, opened 2025-08-08
Issue #1627 fix states for boundary gwt

### LINKED ISSUE #1627: Make ``states_for_boundary`` also work for imod.mf6.GroundwaterTransportModel.clip_box
I just found that ``states_for_boundary`` only works for ``imod.mf6.GroundwaterFlowModel.clip_box``.

We can refactor ``clip_box`` so that we don't need an implementation for each child class of Modflow6Model.

JoerivanEngelen:
Fixes #1627

# Description
- Refactor: Generalize boundary creation so that it will also work with little effort for GroundwaterTransportModel (and GroundwaterEnergy in the future?). 

# Checklist
- [x] Links to correct issue
- [ ] Unit tests were added

### DIFF
--- a/imod/mf6/model.py
+++ b/imod/mf6/model.py
@@ -29,6 +29,9 @@ from imod.common.utilities.schemata import (
 )
 from imod.common.utilities.version import prepend_content_with_version_info
 from imod.logging import LogLevel, logger, standard_log_decorator
+from imod.mf6.chd import ConstantHead
+from imod.mf6.clipped_boundary_condition_creator import create_clipped_boundary
+from imod.mf6.cnc import ConstantConcentration
 from imod.mf6.drn import Drainage
 from imod.mf6.ghb import GeneralHeadBoundary
 from imod.mf6.hfb import HorizontalFlowBarrierBase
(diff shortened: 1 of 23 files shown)


imod/mf6/model.py
```diff
@@ -54,11 +59,72 @@ def pkg_has_cleanup(pkg: Package):
...
+    )
+
+
+def _create_boundary_condition_clipped_boundary(
+    original_model: Modflow6Model,
+    clipped_model: Modflow6Model,
+    state_for_boundary: Optional[GridDataArray],
+    clip_box_args: tuple,
+):
+    unassigned_boundary_original_domain = (
```
Manangka, line 92: Maybe add a comment why this is done?

"Assigning temporary boundary conditions to the original model to the model boundaries that don't have any B.C assigned to them. This makes it possible to identify new boundaries that are created after the clipping (those will be the boundaries without a BC)"

Manangka (approved):

imod/mf6/model.py
JoerivanEngelen: Added:

_Create temporary boundary condition for the original model boundary. This
is used later to see which boundaries can be ignored as they were already
present in the original model. We want to just end up with the boundary
created by the clip._

Issue #278, opened 2022-10-10
upper_active_layer loads all data into memory


Manangka:
In GitLab by @Huite on Oct 10, 2022, 17:24

The first line of `upper_active_layer` is `da = da.load()`.

This doesn't work well for e.g. calculating gxg values for a voxel model, since it will load all times into memory.

JoerivanEngelen:
``upper_active_layer`` is replaced by ``get_upper_active_cells`` which doesn't load data.

PR #1654, opened 2025-08-28
Issue #667 remove duplicate functions

### LINKED ISSUE #667: Duplicate functions: get_upper_active_layer
In GitLab by @JoerivanEngelen on Nov 23, 2023, 17:08

I accidently added a duplicate function ``imod.prepare.layer.get_upper_active_layer``, whereas there already is one under ``imod.prepare.select.layers.get_upper_active_layer``. The latter is a better namespace, but the function in it is not very generic. It does quite a lot of unnecessary xarray operations. I know it is being used though by colleagues, so to solve:

- Move functions under ``imod.prepare.layer.get_upper_active_layer`` to ``imod.prepare.select.layers`` module, so ``imod.prepare.select.layers.get_upper_active_layer`` is replaced.
- If ``imod.select.layers.get_upper_active_layer`` called with dtype not boolean, the old function should be called and a deprecation warning should be thrown.
- I see the old function skips areas without any active cells; this might be behaviour we want in the new functions as well

JoerivanEngelen:
Fixes #667 and #278

# Description
Further reduce code:

- Remove ``upper_active_layer`` function as it is duplicate and not as generic
- Move functions in ``imod.prepare.layer`` to ``imod.select.layers`` module.

# Checklist

- [x] Links to correct issue
- [ ] Unit tests were added


imod/select/layers.py
```diff
@@ -1,41 +1,74 @@
...
+    """
+    layer = active.coords["layer"]
+    upper_active_layer = get_upper_active_layer_number(active)
+    return layer == upper_active_layer
+
+
+@preserve_gridtype
+def get_lower_active_layer_number(active: GridDataArray) -> GridDataArray:
+    """
+    Returns two-dimensional grid of integers with the layer number of the lower
```
Manangka, line 46: Would be good if the documentation of this and the `get_upper_active_layer_number` are consistent

imod/tests/test_select/test_select_layers.py
```diff
@@ -1,117 +1,113 @@
...
+        active_structured[2, :, 5] = False
+        upper_layer_number = ones_like(active_structured.sel(layer=1)).astype(int)
+        lower_layer_number = upper_layer_number + 2
+        upper_layer_number[:, 1] = 2
+        upper_layer_number[:, 3] = 3
+        lower_layer_number[:, 4] = 1
+        lower_layer_number[:, 5] = 2
+        return active_structured, upper_layer_number, lower_layer_number
+
+    def case_unstructured(self, active_unstructured):
```
Manangka, line 59: It's probably me but i find it hard to read what is being done. Can you add a comment stating how you are configuring the grid. The same for the `case_structured`

imod/select/layers.py
```diff
@@ -1,41 +1,74 @@
...
-    da = is_active.layer.isel(layer=is_active.argmax(dim="layer"))
-    da = da.drop_vars("layer")
-
-    # skip where no active cells
-    return da.where(is_active.sum(dim="layer") > 0)
+    active: {DataArray, UgridDataArray}
+        Grid of booleans designating active cell.
+    """
+    layer = active.coords["layer"]
+    # Set nodata to sentinel value to prevent a dtype shift from integer to
```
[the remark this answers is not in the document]
JoerivanEngelen: Ah value like -9999 indicating nan. Numpy doesn't support regular ``np.nan`` for integers, that only works for floats.
In this case I set the sentinel value to ``layer.max() + 1`` or ``layer.min() - 1``, to always be outside the range of layer.

imod/select/layers.py
JoerivanEngelen: Good catch!

imod/tests/test_select/test_select_layers.py
JoerivanEngelen: Fair point, added a comment here:

```python
        # Deactivate some cells in different layers to create a more complex
        # pattern.
```

Issue #996, opened 2024-04-29
Add user guide page on how to use allocation and distribution utilities


JoerivanEngelen:
Since we finished this: https://github.com/Deltares/imod-python/issues/628, we now have the cell allocation and conductance distributing utilities in place. A nice user guide explaining this functionality to uses is still required. 

- Add/create some example data with planar grids containing river/ditch information
- Explain how to allocate cells over layers
- Explain how to distribute conductances over layers
- Plot cross-section to show results, show effects of different methods

PR #1673, opened 2025-09-08
Issue #996 Add Weekly CI job that runs the acceptance tests

### LINKED ISSUE #988: Add weekly/monthly test to CI for large model
We currently don't do any automatic testing on large models. Would be good to process a large model on TeamCity to see if performance recent changes didn't cause any major performance loss and to reassure that iMOD Python can still deal with large, complex datasets. 

The LHM-fresh salt could be used for this. When we have the iMOD5 backwards compatibility, we could also test on LHM fresh.

Manangka:
Fixes #988

# Description
This commit adds a weekly job that runs the acceptance tests.
A main "Test" job is created that is triggered every sunday. This job triggers its dependencies.
For now that is only the "acceptance tests" job. However any future weekly jobs can be attached to it.
I would also make sense to eventually put the "Pixi update" under this job as that is a weekly job as well

# Checklist

- [x] Links to correct issue
- [ ] Unit tests were added

### DIFF
--- a/.teamcity/Templates/AcceptanceTestsTemplate.kt
+++ b/.teamcity/Templates/AcceptanceTestsTemplate.kt
@@ -1,0 +1,46 @@
+package Templates
+
+import jetbrains.buildServer.configs.kotlin.DslContext
+import jetbrains.buildServer.configs.kotlin.Template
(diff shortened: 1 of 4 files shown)


JoerivanEngelen (approved):
2 comments, approving in advance.

PR #1686, opened 2025-09-24
Model split perf

Huite:
For @jdelsman

Sample script:

```python
# %%

import imod

# %%
path = "Z:/tmp/joost/v5_composiet_v1b_v2_nodsp_aut/v5_composiet_v1b_v2_nodsp.toml"
sim = imod.mf6.Modflow6Simulation.from_file(path)
submodel_labels = sim.create_partition_labels(npartitions=24)

# %%

sim_partitions = sim.split(submodel_labels, ignore_time_purge_empty=True)
```

The `sim.split()` takes about 70 seconds here, which seems reasonable (compared to 2 hours before?).

### DIFF
--- a/imod/mf6/boundary_condition.py
+++ b/imod/mf6/boundary_condition.py
@@ -202,12 +202,17 @@ class BoundaryCondition(Package, abc.ABC):
         options = copy(predefined_options)
 
         if not_options is None:
-            not_options = self._get_period_varnames()
+            not_options = []
+            if hasattr(self, "_period_data"):
+                not_options.extend(self._period_data)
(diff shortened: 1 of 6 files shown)


Huite:
Some general thoughts (I've had a little time to mull it over); we a minior optimization in the `from_file` method, which is to simply load all variables that are non-spatial. For these variables, having it dask backed will only increase overhead. This'll also solve the `.item()` issue.

Secondly, in the get_unfiltered_options; we should probably use `is_scalar` as a predicate.

Manangka:
This issue has been addressed in #1693

Manangka:
The second part of this issue has been addressed in #1700

PR #1700, opened 2025-10-20
Issue #1698 Model write optimization

### LINKED ISSUE #1698: Improve write operation for large models

Manangka:
Fixes #1698

# Description
This is part 2 of fixing the performance issues with large model.
In part 1 #1693 the modelsplitter has been optimized. In this PR the focus is on wiring the partitioned model.

As @Huite pointed out in  #1686 the performance bottleneck had to do with the fact that the same package had to be loaded from file multiple times while only a part of the file is actually needed.

After digging around for a while i discovered that this had to do with the fact how we open de the dataset. 
`dataset = xr.open_dataset(path, **kwargs)`

In the line above we don't specify anything chunk related. That has as a result that when you access the dataset the entire file has to be loaded from disk. By simply adding `chunks="auto"` this is no longer the case and a huge performance gain is achieved.

There are some other changes related to setting chunking to auto. There are some parts of the code that don't expect to receive dask arrays. For instance you can use .item() on a dask array. Instead i now use .values[()].

I was also getting some errors when the to_netcdf method were called on the package. All of them had something to do with wrong/unsupported datatypes. In this PR you will find that an encoding is added for float16 types. And that in some packages the from_file method has been updated to ensure that he loaded type is converted to a supported type

An unrelated change but performance wise significant change has been applied to the `_get_transport_models_per_flow_model` method. This method is used to match gwf models to gwt models so that gwfgwt exchanges can be created. This method was doing a full comparison of domains, which is expensive. There is also a method available that does the comparison on domain level. By switching to this method the matching algorithm becomes almost instantaneously.

**NOTE**
This PR has issue #1699 as a base. The base needs to altered to master once that PR is in

**NOTE**
This PR also improves the `dump` method

**NOTE**
some timmings:

image

image

image

- [x] Links to correct issue
- [ ] Unit tests were added

### DIFF
--- a/imod/mf6/wel.py
+++ b/imod/mf6/wel.py
@@ -1073,7 +1073,7 @@ class Well(GridAgnosticWell):
     ) -> pd.DataFrame:
         # Ensure top, bottom & k
         # are broadcasted to 3d grid
-        like = ones_like(active)
+        like = ones_like(active.compute())
         bottom = like * bottom.compute()
         top_2d = (like * top.compute()).sel(layer=1)
         top_3d = bottom.shift(layer=1).fillna(top_2d)


imod/mf6/rch.py
```diff
@@ -166,6 +167,22 @@ def __init__(
...
+        # which are loaded as dtype=object arrays.
+        # This will convert them back to str.
+        vars = [
+            "species",
+        ]
+        for var in vars:
+            if var in instance.dataset:
+                instance.dataset[var] = instance.dataset[var].astype(str)
+
+        return instance
```
JoerivanEngelen: I think this should be part of the pkgbase class. Or BoundaryCondition class, as species can also be part of other packages.

JoerivanEngelen (approved):
I've fixed my comments, thanks for creating this. This indeed speeds up write times considerably for the example!

PR #1706, opened 2025-10-30
Issue #1683: Support zarr and zipstore for dump and from_file methods in mf6 model

### LINKED ISSUE #1683: to_file / dump / etc should support Zarr with ZipStore specifically
Related to #293

Zarr has a number of nice benefits over regular netCDF4. However, it comes with a large downside, which is that the Zarr hierarchy is just files. This makes the files visible to the operating system, which makes moving them around a lot slower. The solution here is to use a 0-compression ZipStore instead. This'll render the zarr store intransparent, which makes e.g. copying significantly easier.

If we support zarr, we should perhaps default to a ZipStore (and also make sure it's easily supported in reading the data back into Python).

I see there's an old issue of mine still open at xarray: https://github.com/pydata/xarray/issues/4076

JoerivanEngelen:
Fixes #1683

# Description
This PR salvages the logic from #1686 to dump and import files from zarr and zip store and adds unit tests.

# Checklist
- [x] Links to correct issue
- [x] Unit tests were added
- [x] **If feature added**: Added feature to API documentation

### DIFF
--- a/docs/api/mf6.rst
+++ b/docs/api/mf6.rst
@@ -81,8 +81,7 @@ Discretization
     StructuredDiscretization.get_regrid_methods
     StructuredDiscretization.regrid_like
     StructuredDiscretization.from_file
-    StructuredDiscretization.to_netcdf
-    StructuredDiscretization.to_zarr
+    StructuredDiscretization.to_file
     StructuredDiscretization.copy
     StructuredDiscretization.clip_box
     StructuredDiscretization.mask
@@ -92,8 +91,7 @@ Discretization
     VerticesDiscretization.get_regrid_methods
     VerticesDiscretization.regrid_like
     VerticesDiscretization.from_file
-    VerticesDiscretization.to_netcdf
-    VerticesDiscretization.to_zarr
+    VerticesDiscretization.to_file
     VerticesDiscretization.copy
     VerticesDiscretization.clip_box
     VerticesDiscretization.mask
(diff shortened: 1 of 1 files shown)


Manangka:
[serializer.zip](https://github.com/user-attachments/files/23256891/serializer.zip)

@JoerivanEngelen I've add a patch with a view changes this PR can benefit from.
Instead of having all the logic of writing files spread over the simulation, model and packages classes I propose consolidating them in 1 single file.
We would end up with 2 files, 1 for the netcdf serialization and 1 for the zarr serialition. This makes the code also easy extendable to support serializing to other formats

Manangka:
When running the unittests i get the following error often:

`C:\dev\imod-python\.pixi\envs\default\Lib\site-packages\_pytest\unraisableexception.py:33: RuntimeWarning: deallocating CachingFileManager(<class 'netCDF4._netCDF4.Dataset'>, 'C:\\Users\\titus\\AppData\\Local\\Temp\\pytest\\pytest-1287\\test_split_dump_netcdf4_0\\split\\tpt_d_0_tpt_d_1.gwtgwt.nc', mode='r', kwargs={'clobber': True, 'diskless': False, 'persist': False, 'format': 'NETCDF4'}, manager_id='bf20bf87-26fd-48ea-a1c7-17925b05b8f5'), but file is not already closed. This may indicate a bug.`

I'm not sure if this has anything to do with your code or that it was already there. Could you verify if this is new or not and create a separate story to fix this if it is exsiting behaviour?

You can reproduce it using:
`pixi run pytest -k test_split_dump[netcdf4]`

PR #1742, opened 2025-12-16
Issue #1738 add ignore time arguments

### LINKED ISSUE #1738: [Performance] - 2 calls to ``purge_empty_packages`` are missing ``ignore_time`` argument
In the calls in:

- _mask_all_packages
- Modflow6Model.clip_box

The ``ignore_time`` argument is missing. Having this option available would help improving performance a lot when dealing with datasets with a lot of stress periods.

JoerivanEngelen:
Fixes #1738

# Description

- Adds missing ignore_time_purge_empty arguments to allow users to ignore time
- Remove underscore of two functions which were used outside the module, as this apparantly is not pythonic

# Checklist

- [x] Links to correct issue
- [x] Unit tests were added

### DIFF
--- a/imod/mf6/simulation.py
+++ b/imod/mf6/simulation.py
@@ -1726,7 +1726,7 @@ class Modflow6Simulation(collections.UserDict, ISimulation):
             timesteps for each package is a costly operation. Therefore, this
             option can be set to True to only check the first timestep. Defaults
             to False.
-            
+
         Examples
         --------
         To mask all models in a simulation, you can use the following code:
--- a/imod/tests/test_mf6/test_mf6_rch.py
+++ b/imod/tests/test_mf6/test_mf6_rch.py
@@ -299,7 +299,7 @@ def test_validate_false():
 def test_ignore_time_validation():
     """
     Create a large recharge dataset with a time dimension. This is to test the
-    performance of the validation when ignore_time_no_data is True. 
+    performance of the validation when ignore_time_no_data is True.
     NOTE: There currently is no easy way to test the opposite (i.e.,
     ignore_time_no_data is False, then catch timeout with an pytest xfail
     marker), because the timeout will terminate the test run with an error when


imod/mf6/model.py
```diff
@@ -724,6 +725,9 @@ def clip_box(
         state_for_boundary : optional, Union[xr.DataArray, xu.UgridDataArray]
             A grids with states that are used to put as boundary values. This
             model will get a :class:`imod.mf6.ConstantHead`.
+        ignore_time_purge_empty : bool, default False
+            Whether to ignore time dimension when purging empty packages. Can
+            improve performance when clipping models with many time steps.
```
Manangka: I think this explanation can use some improvement. 
Maybe elaborate more on the fact that after clipping packages can become empty e.g. ...
An package is considered empty when ...
This boolean determines if the time dimension is taken into account when determining empty packages

imod/tests/test_mf6/test_mf6_mask_simulation.py
```diff
@@ -9,13 +9,15 @@
 from imod.tests.fixtures.mf6_modelrun_fixture import assert_simulation_can_run
 
 
+@pytest.mark.parametrize("ignore_time", [True, False])
```
Manangka: In all the test it doesnt matter what the flag does
It would be good to have some tests where the outcome is different depending on the flag

Manangka (changes requested):

imod/tests/test_mf6/test_mf6_mask_simulation.py
JoerivanEngelen: Indeed, we already have a test that do this:

- ``test_purge_empty_package__ignore_time`` in ``imod/tests/test_mf6/test_mf6_model.py``

I also tried adding a test to test in ``imod/tests/test_mf6/test_mf6_rch.py`` next to the performance test ``test_ignore_time_validation`` to test the opposite, this failed however, as there is no easy way to catch fails due to timeouts (something with an xfail) and the test errors instead of fails due to dask when the timeout kicks in. I added a note instead so that I do not waste on this again in the future ;-). This test would only really test whether the performance test ```test_ignore_time_validation`` is not unexpectedly returning false positives though, which is a nice-to-have. ``test_purge_empty_package__ignore_time`` already tests if the data handling is correct.

I think spending a lot of time on adding more tests in ``test_mf6_mask_simulation`` for this is a bit of a time waster.

Manangka (approved):

PR #1771, opened 2026-02-09
Issue #1768 pandas3 support

### LINKED ISSUE #1768: Support pandas 3.0
The update to pandas 3.0 breaks the unittests and examples. See: https://github.com/Deltares/imod-python/pull/1767 

From a quick glance I see some results of asserts change only very slightly and fail (e.g. less trailing zeros for Timestamps ``"2013-04-29T22:00:00.000000"`` instead of ``"2013-04-29T22:00:00.000000000"``). I don't think that should be causing problems for users.

However, there are also appear to be some stricter type checks in pandas 3.0 causing quite some failures, which might cause problems for users as well.

JoerivanEngelen:
Fixes #1768 

# Description
Fixes broken pieces of the code when updating to pandas 3, which [has the following major changes](https://pandas.pydata.org/community/blog/pandas-3.0.html) which also affect iMOD Python somewhat:

- _Dedicated string data type by default: string columns are now inferred as the new str dtype instead of object, providing better performance and type safety_
- _Consistent copy/view behaviour with Copy-on-Write (CoW) (a.k.a. getting rid of the SettingWithCopyWarning): more predictable and consistent behavior for all operations, with improved performance through avoiding unnecessary copies_
- _New default resolution for datetime-like data: no longer defaulting to nanoseconds, but generally microseconds (or the resolution of the input), when constructing datetime or timedelta data (avoiding out-of-bounds errors for dates with a year before 1678 or after 2262)_

Because of these changes, some parts of the code had to be modified slightly to get the tests to work. I think with the latter change we could simplify the code base a lot, as all logic to deal with datetimes going beyond the year 2262 and prior to 1678 wouldn't be necessary anymore. This was quite a headache in the past, and I think the choice for microseconds by default will make our lives significantly easier. [I created an issue for this](https://github.com/Deltares/imod-python/issues/1773).

In detail this PR alters the following:

- Pin pandas version to 3.* in pixi.toml
- Update setup and asserts in quite some tests to work with pandas 3.0. Most of these tests will not work with pandas 2.0 anymore, as we now assert for pandas 3.0 behavior
- Update examples to work with pandas 3.0
- Fix statements where a view was altered (e.g. altering ``.values`` attribute), this is now forbidden by pandas
- Update to new string handling, sometimes I had to internally enforce object dtype to get the old behavior. In other cases we could check for strings by checking ``dtype=="object"`` anymore. Luckily there is ``pd.api.types.is_string_dtype``, which both works for pandas 2 as well as pandas 3.
- Add unittest for ``to_pandas_datetime_series`` to check whether pandas behavior is similar in this regard.

# Checklist

- [x] Links to correct issue
- [x] Unit tests were added


pixi.toml
```diff
@@ -88,7 +88,7 @@ netcdf4 = "*"
 numba = ">=0.50"
 numpy = "*"
 pandamesh = "*"
-pandas = "*"
+pandas = "3.*"
```
Manangka: Should this be equal or bigger than 3?

Manangka (approved):

PR #1803, opened 2026-04-09
Speed up TeamCity build

### LINKED ISSUE #1804: Speed up Teamcity build
The TeamCity build takes longer and longer to complete.
Two possible tracks are being followed to resolve this.

The first one is done by @JoerivanEngelen. He is looking into identifying bottlenecks in the units tests.
The second one is done by @Manangka. He is looking into identifying bottlenecks in the TeamCity build

Manangka:
Fixes #1804

In an effort to speed up the CI-pipeline 2 changes have been made:
- After some investigation a bottleneck has been identified during the CI-pipeline. 
The CI build steps are run in docker containers. In this containers a volume is  mounted which contains the checkout repository.  This is the way TeamCity works.
When installing the pixi environment the environment the environment is installed into this mounted volume. By moving the environment out of the mounted volume to another location in the container a noticeable speedup is achieved.
- The editable installation of imod has been changed from a task to a pypi-dependency. This way it happens at the same time as when the pixi environment is being installed, which results in a slight speedup

**Unrelated**
The mypy daemon kept crashing in vscode. The reason is that it encountered some partial types when analyzing pyvista.
I've disabled following that import. Now its possible to use the daemon i.c.w. the mypy extension in vscode, which should give you faster feedback when new mypy errors popup

# Description

# Checklist

- [ ] Links to correct issue
- [ ] Unit tests were added

### DIFF
--- a/.teamcity/Templates/AcceptanceTestsTemplate.kt
+++ b/.teamcity/Templates/AcceptanceTestsTemplate.kt
@@ -30,6 +30,9 @@ object AcceptanceTestsTemplate : Template({
             workingDir = "imod-python"
             scriptContent = """
                 SET PATH=%%PATH%%;%system.teamcity.build.checkoutDir%\modflow6
+
+                pixi config set --local detached-environments "C:\pixi_envs"
+                pixi install --environment user-acceptance --frozen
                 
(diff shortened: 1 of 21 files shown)


JoerivanEngelen:
Nice!

JoerivanEngelen (approved):

PR #1806, opened 2026-04-13
Enforce MyPy "disallow any generics" rule

### LINKED ISSUE #1807: Enforce MyPy "disallow any generics" rule

Manangka:
Fixes #1807 

# Description
This PR fixes the mypy errors when enabling the "disallow any generics"
The majority of the changes are type hints and this no functional  changes.
The exceptions are in:
- simulation.py A key is converted to a string
- msw/model.py‎ Some none checks are added

# Checklist

- [ ] Links to correct issue
- [ ] Unit tests were added

### DIFF
--- a/imod/mf6/model.py
+++ b/imod/mf6/model.py
@@ -6,7 +6,7 @@ import inspect
 import pathlib
 import warnings
 from pathlib import Path
-from typing import Any, List, Optional, Tuple, Union
+from typing import Any, List, Optional, Tuple, Union, cast
 
 import cftime
 import jinja2
@@ -90,7 +90,7 @@ def _create_boundary_condition_clipped_boundary(
     original_model: Modflow6Model,
     clipped_model: Modflow6Model,
     state_for_boundary: Optional[GridDataArray],
-    clip_box_args: tuple,
+    clip_box_args: tuple[Any, ...],
 ) -> Optional[StateType]:
     # Create temporary boundary condition for the original model boundary. This
     # is used later to see which boundaries can be ignored as they were already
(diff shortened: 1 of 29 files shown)


imod/mf6/model.py
```diff
@@ -569,7 +569,7 @@ def _write(
                         write_context=pkg_write_context,
                     )
                 elif issubclass(type(pkg), imod.mf6.HorizontalFlowBarrierBase):
-                    mf6_hfb_ls.append(pkg)
+                    mf6_hfb_ls.append(cast(HorizontalFlowBarrierBase, pkg))
```
JoerivanEngelen: Strange that this has to be cast, wouldn't ``isinstance`` be more fitting here instead of ``issubclass``? Perhaps the casting is not necessary here?

imod/msw/model.py
```diff
@@ -317,6 +317,8 @@ def _write_simulation_settings(self, directory: Path) -> None:
 
         # Add IdfMapping settings
         idf_key = self.get_pkgkey(IdfMapping)
+        if idf_key is None:
```
JoerivanEngelen: Are these checks really necessary btw? We could adapt ``get_pkgkey`` to throw the error there instead? There is a check in MetaSWAP to check if all required packages are present in the model, this check is called before writing the simulation settings.

JoerivanEngelen (approved):
Thanks for fixing this, I have some questions. Approving in advance

imod/mf6/model.py
Manangka: You're right. Replacing it by `isinstance` fixed the issue without having to resort to casting

imod/msw/model.py
Manangka: As discussed, these changes don't make the code easier to read and are also not necessary since it is ensured elsewhere that there can be no None types.
Therefor i reverted these changes and instead added a comment to ignore the mypy errors

PR #1824, opened 2026-04-23
Issue #1823 regrid imod5 cap data

### LINKED ISSUE #1823: [FEATURE] - Regrid iMOD5 data for MetaSWAP models as well.
**Is your feature request related to a problem? Please describe.**
The development of the iMOD5 support, I neglected automatic regridding of the MetaSWAP data in the dbase, as it was not necessary for the LHM. This now has to be implemented for the regional models.

JoerivanEngelen:
Fixes #1823

# Description
Regrid iMOD5 CAP data immediately to MODFLOW6 discretization in ``MetaSwapModel.from_imod5_data`` .  I'm doing this regridding prior to calling the individual ``from_imod5_data`` methods of each MetaSwap package as we can then be sure that all grids in the iMOD5 CAP data are on to the right grid (which is not necessarily the case with the data in projectfile), before other grid operations like ``get_cell_area_from_imod5_data`` are done. Furthermore, the keys in the CAP data differ from the variable names in packages. 

This differs slightly from the approach in the MODFLOW6 module, where regridding is done in ``Package.from_imod5_data``.

Fixes the following:
- Regrid CAP data
- Let ``_regrid_array`` return a DataArray without any dimensions instead of a numpy array if_scalar is True, this is consistent with the type annotation, and has the advantage that these scalars can be broadcasted without having to do additional type conversions in the masking operation later. 
- Boyscouting: Fix typo in "conductivity_factor" key that is used mostly internally

# Checklist

- [x] Links to correct issue
- [x] Unit tests were added

### DIFF
--- a/imod/msw/utilities/imod5_converter.py
+++ b/imod/msw/utilities/imod5_converter.py
@@ -1,12 +1,18 @@
+from typing import Optional
+
 from xarray.core.utils import is_scalar
 
 from imod.common.constants import MaskValues
+from imod.common.utilities.regrid import _regrid_package_data
 from imod.logging import LogLevel, logger
 from imod.mf6 import StructuredDiscretization
+from imod.msw.regrid.regrid_schemes import CapDataRegridMethod
 from imod.msw.utilities.common import concat_imod5
 from imod.msw.utilities.mask import MetaSwapActive
-from imod.typing import GridDataArray, GridDataDict
+from imod.typing import GridDataArray, GridDataDict, Imod5DataDict
 from imod.typing.grid import ones_like
+from imod.util.dims import drop_layer_dim_cap_data
+from imod.util.regrid import RegridderWeightsCache
 from imod.util.spatial import get_cell_area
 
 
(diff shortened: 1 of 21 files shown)


imod/msw/utilities/imod5_converter.py
```diff
@@ -116,3 +122,33 @@ def has_active_scaling_factor(imod5_cap: GridDataDict):
         )
 
     return not scaling_factor_inactive
+
+
+def regrid_imod5_data(
+    imod5_data: Imod5DataDict,
+    target_dis: StructuredDiscretization,
+    regridder_types: Optional[CapDataRegridMethod] = None,
```
Manangka: Instead of this being optional you can assign it a default value of ` CapDataRegridMethod()`. That way you can remove the `if None` statement at line 147

Manangka (approved):

imod/msw/utilities/imod5_converter.py
JoerivanEngelen: Yes, I did it this way to keep things consistent with how ``regridder_types`` are provided in the mf6 module. Looking at the mf6 module, I think there with the current setup it is the best way to do things, but here it isn't really necessary. I'll provide the default value of ``CapDataRegridMethod()`` to ``MetaSwapModel.from_imod5_data``.

PR #1833, opened 2026-05-06
Issue #1810 fix weird gen warning

### LINKED ISSUE #1810: [Bug] - Confusing warning message when loading messy GEN files
**Bug description**
I'm loading messy GEN files from IBRAHYM, but it results in an incorrect warning:

``UserWarning: Inconsistent IPF: header states 2 columns, first line contains 3``

This stems from the ``imod.formats.ipf._infer_delimwhitespace`` helper function from the IPF module being used in the GEN module. 
 
- The following needs to be done: create a small ``imod.formats.common`` module where useful helper functions can be shared
- Add an argument so that the warning message text "IPF" or "GEN" can be set.

JoerivanEngelen:
Fixes #1810 

# Description
This fixes confusing warning about incorrect IPF headers when loading GEN files. 
- Move ``_infer_delim_whitespace`` to a ``imod.formats.common.py`` module
- Throw warning only in IPF module

# Checklist

- [x] Links to correct issue
- [x] Unit tests were added


imod/formats/common.py
```diff
@@ -0,0 +1,30 @@
+import csv
+
+
+def infer_delimwhitespace(line: str, ncol: int):
```
Manangka: This method does 2 things. From the name it isn't clear that it also does a column check.
You could make the name more generic and return a namedtuple. 

The method signature than also contains the return type which makes it clearer what it does

```
class DelimiterInfo(NamedTuple):
    is_whitespace: bool
    has_expected_cols: bool


def infer_delimiter_info(line: str, ncol: int) -> DelimiterInfo:
    """
    Infer whether the line is delimited by whitespace or commas, and whether
    the number of comma-delimited columns matches the expected count.

    Parameters
    ----------
    line : str
        The line to analyze.
    ncol : int
        The expected number of columns if line is delimited by commas.

    Returns
    -------
    DelimiterInfo
        is_whitespace : bool
            Whether the line is delimited by whitespace.
        has_expected_cols : bool
            Whether the line has the expected number of columns if delimited by commas.
    """
    n_elem = len(next(csv.reader([line])))
    if n_elem == 1:
        return DelimiterInfo(is_whitespace=True, has_expected_cols=True)
    elif n_elem == ncol:
        return DelimiterInfo(is_whitespace=False, has_expected_cols=True)
    else:
        return DelimiterInfo(is_whitespace=False, has_expected_cols=False)
```

On the calling side you can then use the named tuple:
```
delimiter_info = infer_delimiter_info(line, ncol)`
delimiter_info.is_whitespace
delimiter_info.has_expected_cols
```
or directly unpack it:
`has_whitespace, has_expected_cols = infer_delimiter_info(line, ncol)`

imod/tests/test_formats/test_format_common.py
```diff
@@ -0,0 +1,10 @@
+from imod.formats.common import infer_delimwhitespace
+
+
+def test_infer_delimwhitespace():
```
Manangka: This test assert many things.
Consider using test cases to make it clearer what you're testing

```
@pytest.mark.parametrize(
    "line, ncol, expected",
    [
        ("1 2 3", 3, (True, True)),    # whitespace-delimited, correct ncol
        ("1 2 3", 2, (True, True)),    # whitespace-delimited, wrong ncol: ncol is ignored
        ("1\t2\t3", 3, (True, True)),  # tab-delimited (csv.reader sees 1 element -> whitespace branch)
        ("1,2,3", 3, (False, True)),   # comma-delimited, correct cols
        ("1, 2, 3", 3, (False, True)), # comma-delimited with spaces, correct cols
        ("1,2,3", 4, (False, False)),  # comma-delimited, wrong number of cols
        ("1 2,3", 3, (False, False)),  # mixed delimiters, unexpected col count
    ],
)
def test_infer_delimiter_info(line, ncol, expected):
    assert infer_delimiter_info(line, ncol) == expected
```

imod/formats/common.py
Manangka: `infer_line_delimiter_info` could also be a good name with the named tuple then being `LineDelimiterInfo`

JoerivanEngelen: Yes, that clarifies the code a lot. I'll update this!

imod/tests/test_formats/test_format_common.py
JoerivanEngelen: Right, I've separated the cases into a separate class for pytest_cases, so that it also looks good in the VSCode test explorer.

Manangka (approved):

Issue #1682, opened 2025-09-22
`.values[()]` in _get_unfiltered_pkg_options should be `.item()`?


Huite:
In fetching options, `.values[()]` is used:

```python
    def _get_unfiltered_pkg_options(
        self, predefined_options: dict, not_options: Optional[list] = None
    ):
        options = copy(predefined_options)

        if not_options is None:
            not_options = []
            if hasattr(self, "_period_data"):
                not_options.extend(self._period_data)
            if hasattr(self, "_auxiliary_data"):
                not_options.extend(self._auxiliary_data.keys())

        for varname in self.dataset.data_vars.keys():  # pylint:disable=no-member
            if varname in not_options:
                continue
            v = self.dataset[varname].values[()]
            options[varname] = v
        return options
```

Unless I'm mistaken, options should always be scalar. So I think `.item()` should be used instead.

`.item()` has the strong advantage of erroring on non-scalar contents. At the moment, the concentration array is fully loaded into memory due to #1681; this would not have happened if `.item()` is used here.

Huite:
I've just tried it with `.item()` and it's clear that the `not_options` set is incomplete.

It goes through:

```
concentration_boundary_type
print_input
print_flows
save_flows
observations
repeat_stress
```

Of which observations and repeat_stress aren't option entries either!

Anyway, I think the best idea is to use .item and make sure no unwanted variables are seen there.

JoerivanEngelen:
If I remember correctly, I did this because this was necessary for dask arrays, which returns something different for scalars when you call .item() instead of .values[()]. I see @Manangka also ran into this in PR https://github.com/Deltares/imod-python/pull/1700.

JoerivanEngelen:
Also somewhat related to https://github.com/Deltares/imod-python/issues/281

Huite:
> If I remember correctly, I did this because this was necessary for dask arrays, which returns something different for scalars when you call .item() instead of .values[()]. I see [@Manangka](https://github.com/Manangka) also ran into this in PR [#1700](https://github.com/Deltares/imod-python/pull/1700).

Hmmm, I guess I experienced that too because I was getting errors from the model that was loaded from zarr storage; I worked around it by calling `.compute` or `.load` on all scalar variables. There's an xarray issue for it: https://github.com/pydata/xarray/issues/7916

Unfortunately the dask PR has not been looked at for a long time: https://github.com/dask/dask/pull/3630

Huite:
[...]

PR #1846, opened 2026-05-27
Issue #1830 Only call item on DataArrays of size 1 computed in memory.

### LINKED ISSUE #1830: [Bug] - Dask arrays don't support item()
**Bug description**
@jdelsman ran into this.

Dask arrays do not support the ``item`` method.
In his case, he could fix this by commenting out the call to ``item``:

```python
def is_empty_dataarray(da: Any) -> bool:
    return isinstance(da, xr.DataArray) and da.isnull().all()#.item()
```

I have to backtrack again why the call to ``.item`` was necessary here: Was it introduced because it is necessary in some cases, or was it because of a mypy error?

We also need to test more with dask array fixtures to catch these bugs.

Related to #281 and #1682

JoerivanEngelen:
Fixes #1830

# Description
Fix issue where the code could break, as dask arrays are not supported by ``xr.DataArray.item()``. This was for example an issue when settings are loaded by ``imod.mf6.Modflow6Simulation.from_file``.

This PR replaces selected calls to ``xr.DataArray.item()`` or ``xr.DataArray.values[()]`` with calls to ``enforce_scalar``, which does a size check first, then calls a compute to get the item.

Futhermore, the scope of ``enforce_scalar`` is reduced somewhat: it now purely works on DataArrays, instead of a mix of numpy arrays and DataArrays. This caused that I had to remove the call to enforce_scalar in ``reading.py``, but there was already a size check before this call, so ``enforce_scalar`` didn't provide any added value here.

Additional boyscouting: add some missing type annotation to Lake in ``imod.mf6.lak.py``.

# Checklist

- [x] Links to correct issue
- [x] Unit tests were added

### DIFF
--- a/imod/mf6/npf.py
+++ b/imod/mf6/npf.py
@@ -10,6 +10,7 @@ from imod.common.utilities.dataclass_type import (
 from imod.common.utilities.regrid import (
     _regrid_package_data,
 )
+from imod.common.utilities.value_filters import enforce_scalar
 from imod.logging import init_log_decorator
 from imod.mf6.package import Package
 from imod.mf6.regrid.regrid_schemes import (
@@ -35,13 +36,11 @@ def _dataarray_to_bool(griddataarray: GridDataArray) -> bool:
     if griddataarray is None or griddataarray.values is None:
         return False
 
-    if griddataarray.values.size != 1:
-        raise ValueError("DataArray is not a single value")
-
     if griddataarray.values.dtype != bool:
         raise ValueError("DataArray is not a boolean")
 
-    bool_value = cast(bool, griddataarray.values.item())
+    bool_value = cast(bool, enforce_scalar(griddataarray))
+
     return bool_value
 
 
--- a/imod/mf6/npf.py
+++ b/imod/mf6/npf.py
@@ -1,4 +1,4 @@
-from typing import Optional, cast
+from typing import Optional
 
 import numpy as np
 import xarray as xr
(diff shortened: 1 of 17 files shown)


JoerivanEngelen:
I thought I could also easily fix https://github.com/Deltares/imod-python/issues/1682 in this PR, but this requires additional fixes. I'll do that in an additional PR while I'm at it.

imod/mf6/npf.py
```diff
@@ -35,13 +36,11 @@ def _dataarray_to_bool(griddataarray: GridDataArray) -> bool:
...
-    if griddataarray.values.size != 1:
-        raise ValueError("DataArray is not a single value")
-
     if griddataarray.values.dtype != bool:
         raise ValueError("DataArray is not a boolean")
 
-    bool_value = cast(bool, griddataarray.values.item())
+    bool_value = cast(bool, enforce_scalar(griddataarray))
+
     return bool_value
```
LuukBlom: `.values` loads the entire np array into memory, which is we are trying to avoid when a data array is lazyloaded. 

We should try to find a way to determine the size / dtype etc without loading it fully.
Probably something like `GridDataArray.dtype` and `GridDataArray.size` should work?

LuukBlom (commented):
LGTM.

Good to note that accessing `.values` (or the new `.to_numpy()`) will still load the whole dataarray into memory if it is a lazy dask array.
So its a good idea to check for this pattern across the codebase.

imod/mf6/npf.py
JoerivanEngelen: Gotcha, this was something that can be immediately cleaned up as well. Fixed that

LuukBlom (approved):

PR #1847, opened 2026-05-28
Issue #1682 only scalars in package options

### LINKED ISSUE #1681: _get_period_varnames does not work as documented
This function:

```python

    def _get_period_varnames(self) -> list[str]:
        """
        Get variable names for transient data of this package.

        Returns
        -------
        list[str]
            List of variable names that are used for transient data in this
            package.

        Examples
        --------
        To get the variable names for transient data in a package, e.g. a River
        package:

        >>> river = imod.mf6.River.from_file("river_with_concentration.nc")
        >>> river._get_period_varnames()
        >>> # prints: ['stage', 'conductance', 'bottom_elevation', 'concentration']
        """
        result = []
        if hasattr(self, "_period_data"):
            result.extend(self._period_data)
        if hasattr(self, "_auxiliary_data"):
            result.extend(get_variable_names(self))

        return result
```

Does not return what the docstring mentions: `['stage', 'conductance', 'bottom_elevation', 'concentration']`.

Instead of "concentration", it returns the name of species. This is due to how get_variable_names works, and the fact that `_auxiliary_data` is a dict.

This has rather bad effects; in `_get_unfiltered_pkg_options`, not_options should be skipped. Currently, it will read the entire concentration array into memory as `.values[()]` is called.

However, changing the function to:

```python
        if hasattr(self, "_auxiliary_data"):
            result.extend(self._auxiliary_data.keys())
```

Seems to result in other errors, so it looks like other functionality relies on the incorrect behavior!

### LINKED ISSUE #1682: `.values[()]` in _get_unfiltered_pkg_options should be `.item()`?

JoerivanEngelen:
Fixes #1682 and #1681

This can be merged after https://github.com/Deltares/imod-python/pull/1846

# Description
Fixes an issue where big dask arrays were silently loaded into memory when looking for options. Calling ``enforce_scalar`` will now prevent this from happening. 

Furthermore, we were misusing the method ``self._get_period_varnames`` to get the right varnames. The docstring of this method was wrong, which caused probably the misuse. 

Last, because I had to change which ``period_var`` was used to compute maxbound for the Mf6HorizontalFlowBarrier, we fixed a bug where the well package computed an erronous maxbound. Previously it took ``cellid``, but this has an additional dimension, resulting in maxbounds which were thrice too large for structured grids, and twice too large for unstructured grids.

I had to exclude "concentration" and "repeat_stress" explicitly in ``_get_unfiltered_pkg_options()``. This feels a bit arbitrary, but I couldn't find a good place and clear name to store these. 

Further cleanup:
- Make code of Mf6HorizontalFlowBarrier more consistent with the rest.

# Checklist

- [x] Links to correct issue
- [ ] Unit tests were added


LuukBlom:
LGTM.

Though I'm not familiar with the code, I'd say all of these changes make sense

JoerivanEngelen:
There still seem to be some issues with this PR inferring from the tests:

- The CHD package in a test has a maxbound of 0, this shouldn't be the case
- Loading a HFB from a dumped model fails now
- The roundtrip for MetaSWAP fails on an equality check. Needs some investigation what happened here. Possibly a None changed to np.nan.
- Splitting and dumping causes an error, because the modelnames list has 0 length for some reason

JoerivanEngelen:
OK, I think I have fixed the broken tests: Probably the result at a poor merge conflict fix where I accidentily removed and renamed utility functions.

PR #1849, opened 2026-06-01
Fix AttributeError in facebudget when front is omitted (fixes #1097)

### LINKED ISSUE #1097: None access in budget.py
I was playing around with Coverity Scan and it gave me two results, which is a super nice score in the first place!

> CID 429696: (#1 of 1): Attribute/item access or function call before check for None or undefined (REVERSE_INULL) check_after_deref: Null-checking front suggests that it may be null, but it has already been dereferenced on all paths leading to the check. 

In budget.py:292, we see the use of `front`, but `front` can be `None` here: `for itime in range(front.coords["time"].size):`.
And within the `for`, there is again a check to see if `front` is `None`.

gaoflow:
### Summary

`imod.evaluate.facebudget` documents and enforces that **at least one** of
`front`, `lower`, `right` must be supplied (it raises `ValueError` if all three
are `None`). However two code paths dereferenced `front` unconditionally, so
omitting `front` crashed:

```python
def facebudget(budgetzone, front=None, lower=None, right=None, netflow=True):
    ...
    if "time" in dims:
        for itime in range(front.coords["time"].size):   # front may be None
    ...
    else:                                                 # no zones defined
        dask_front = dask.array.full(front.shape, ...)    # front may be None
        dask_lower = dask.array.full(lower.shape, ...)    # lower may be None
        dask_right = dask.array.full(right.shape, ...)    # right may be None
```

So a perfectly valid call such as `facebudget(budgetzone, lower=lower)` raised:

```
AttributeError: 'NoneType' object has no attribute 'coords'   # transient case
AttributeError: 'NoneType' object has no attribute 'shape'    # no-zones case
```

This is exactly the null-dereference Coverity flagged in #1097
(`REVERSE_INULL`: `front` is dereferenced on all paths *before* the
`if front is not None` guard inside the loop). The other location named in
#1097 (`package.py:292`) no longer exists — that code was refactored into
`imod/evaluate/budget.py`, which is the surviving occurrence fixed here.

### Fix

Pick the first supplied DataArray as a reference for the time axis and the
output shape. All three inputs are already validated to share `budgetzone`'s
spatial shape, so any of them is a valid reference:

```python
reference = next(da for da in (front, lower, right) if da is not None)
...
for itime in range(reference.coords["time"].size):
...
dask_front = dask.array.full(reference.shape, np.nan, chunks=chunks)
dask_lower = dask.array.full(reference.shape, np.nan, chunks=chunks)
dask_right = dask.array.full(reference.shape, np.nan, chunks=chunks)
```

### Reproduction / verification

```python
import numpy as np, xarray as xr, imod
da = xr.DataArray(np.zeros((3, 2, 1)),
                  {"layer": [1, 2, 3], "y": [0.5, 1.5], "x": [0.5]},
                  ("layer", "y", "x"))
lower = xr.full_like(da, 0.0); lower[:2] = 1.0
zone = xr.full_like(da, np.nan); zone[:2] = 1
imod.evaluate.facebudget(zone, lower=lower)   # before: AttributeError; after: works
```

Added `test_omit_front` and `test_omit_front_transient` to
`imod/tests/test_evaluate/test_budget.py`. The transient test exercises **both**
crash sites (the time loop and the no-zones branch) via a partial zone and an
all-inactive zone. Verified locally: both tests fail on `master`
(`AttributeError`) and pass with this change; all pre-existing budget tests
(9) still pass.

Fixes #1097.

### DIFF
--- a/docs/api/changelog.rst
+++ b/docs/api/changelog.rst
@@ -74,6 +74,10 @@ Fixed
   settings were specified as dask array, which could happen when loading a model
   lazily with :meth:`imod.mf6.Modflow6Simulation.from_file` and not
   computing the data before writing.
+- Fixed bug where :func:`imod.evaluate.facebudget` raised an error when the
+  ``front`` budget was left out, even though you only need to provide one of
+  ``front``, ``lower`` or ``right``. Leaving out ``front`` now works as
+  described in the documentation.
 
 Changed
 ~~~~~~~
--- a/imod/evaluate/budget.py
+++ b/imod/evaluate/budget.py
@@ -269,6 +269,11 @@ def facebudget(budgetzone, front=None, lower=None, right=None, netflow=True):
             if da_shape != shape:
                 raise ValueError(f"Shape of {daname} does not match budgetzone")
 
+    # At least one of front, lower, right is guaranteed to be provided (checked
+    # above), but any of them may be omitted. Use the first one that is present
+    # as the reference for the time axis and the array shape.
+    reference = next(da for da in (front, lower, right) if da is not None)
+
     # Determine control surface
     face = _outer_edge(budgetzone)
     # Convert indices array to dask array, otherwise garbage collection gets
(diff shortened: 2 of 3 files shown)


JoerivanEngelen:
@gaoflow 

Looks good to me, could you also add your fix to the changelog? Write about it with users in mind, who are not software developers.

binggao1230:
Done — added a Fixed entry to the changelog under [Unreleased], phrased for users: it notes that `facebudget` previously errored when you left out `front`, even though only one of `front`/`lower`/`right` is required, and that omitting `front` now works as documented.

PR #1883, opened 2026-07-22
Issue #1835 sprinkling from ipf

### LINKED ISSUE #1835: [FEATURE] - ``MetaSwapModel.from_imod5_data`` sprinkling from IPF files
**Is your feature request related to a problem? Please describe.**
We currently only support defining sprinkling wells for MetaSWAP from IDF files (IARMWP=0). In this version:

- the ART grid defines the artificial recharge type (1=groundwater, 2=surface water) 
- the ARL grid containing layer numbers where well is located
- the ARC grid defines the artificial recharge capacity.

For the NHI, defining sprinkling from IPF files is increasingly used, so we therefore also have to support this (IARMWP=1).

Refinement
-------------

The following needs to be supported:

- The ART grid contains the IDs to which wells in the IPF file are coupled.
- The ARL is an IPF file that has following columns (which is loaded as a pandas Dataframe):
   - ``x``: x-location of well.  In case the location of the point is outside the model or does not match with existing location (ART), the source of the irrigation water is  automatically from an external source (surfacewater) and equal to ART=2. 
   - ``y``: y-location of well
   - ``ilay``: layer where well is placed. If the layer number exceed the maximal number of model layers, the assigned to layer is set to this maximal model layer. If the assigned layer number is zero, the source of the artificial recharge becomes surface water (ART=2). If the thickness of the assigned model layer is zero, iMOD tries to reposition the well upwards in a model layer with a transmissivity of 0.1 m2/d, or if that doesn't work downwards. If that doesn't work as well, the well is removed entirely.
   - ``id``: id number in ART grid to connect wells to
   - ``capacity``: artificial recharge capacity (mm/d).
- The ARC grid is not used

There is a notable difference with the existing implementation is that the setup with IPF files allows to connect multiple sprinkling wells to a single svat. The content of the IPF file will be loaded as a pandas Dataframe, but there is also grid information required, which is stored in xarray DataArrays. This requires an entirely different render method, so this requires either some dispatching, or perhaps an entirely different object?

The following data processing needs to be done roughly:

- Map IDs from ART grid to svats
- Find sprinkling well location in svat grid and get corresponding svat, if outside grid, set to external source (surface water)
- Allocate wells if ILAY = 0
- Convert artificial recharge capacity from mm/d to m3/d.

Approach
-----------

I think the following approach is probably the most convenient:

- Add support in the LayeredWell and Well package for an additional aux var with aux dimension, something like ``{"sprinkling": "sprinkling_var"}``. This can store the ``ART_id`` and ``capacity`` columns. All the well assignment logic will be done in these objects.
- Split based on layer numbers, wells with layer=0 are separated into a ``Well``, whereas those with layer>0 are assigned to ``LayeredWell``. 
- The logic to reassign wells to different layers for ``LayeredWell`` can be added to ``LayeredWell.cleanup``.  
- This needs to be forwarded to ``Mf6Wel`` package, which needs to store them as aux vars as well in its dataset
- Add an additional ``SprinklingPoints`` object or something, containing only one variable, the ART grid.
- ``SprinklingPoints.render`` uses the aux vars of ``Mf6Well``.

This means that sprinkling well data will be provided to the ``Well`` and ``LayeredWell`` package in a slightly complicated manner, as we need to use explicitly set numpy dtypes to preserve dtype. :

```python
import xarray as xr
import numpy as np

dtype = np.dtype([("id_ART", np.int64), ("capacity", np.float32)])
a = np.array([(1, 1.0), (2, 2.0), (3, 3.0)], dtype=dtype)

da = xr.DataArray(a, coords={"sprinkling_var": [1, 2, 3]}, dims=["sprinkling_var"])
ds = xr.Dataset({"sprinkling": da})
```

Still to figure out: 
- What to do with wells that were clipped away? These need to be converted to surface water routing in some way...
- How good is the current auxvar logic tested?

JoerivanEngelen:
Fixes #1835

# Description
Adds experimental support for ``SprinklingPoints`` class

Changes the following things:

- Add ``SprinklingPoints`` class, to specify a n:1 coupling of grid cells to points.
- Rename ``Sprinkling`` class to ``SprinklingGrid`` class, deprecate ``Sprinkling`` class. I had to keep the API stable as [``primod`` relies on the existence of the ``Sprinkling`` class](https://github.com/Deltares/imod_coupler/blob/c9e838fc93549714fa671ae653aeaa32d2203267/pre-processing/primod/driver_coupling/metamod.py#L86).
- As ``primod`` relies on a call to ``MetaSwapModel.get_pkg_type(Sprinkling)``, I had to modify the code somewhat, so that it returns the key of the sprinkling package regardless whether it is a ``SprinklingPoints`` or a ``SprinklingGrid``
- Add ``LayeredWell.from_imod5_cap_data`` to construct a MODFLOW6 well package from the iMOD5 data under the CAP header.
- Pass on well ``id`` to the ``Mf6Wel`` class' dataset. This is used to map MODFLOW6 wells to MetaSWAP wells. It can furthermore be used in the future to add as auxiliary variable in well files in the future to have some metadata in the MODFLOW6 well file.
- Refactor of the test module ``imod/tests/test_msw/test_sprinkling.py``: Separate the test cases from the tests. 
- Boyscouting: Replace some calls of ``.values`` to ``to_numpy()``
- Boyscouting: Replace some calls of ``.values[()]`` to the safer ``enforce_scalar_data``

This still has to be tested whether it works with the Peelvenen model, I'll do that in follow-up issue #1908 , where I'll also add a larger regression/user_acceptance test for our Weekly tests.

# Checklist

- [x] Links to correct issue
- [x] Unit tests were added
- [x] **If feature added**: Added feature to API documentation


imod/mf6/mf6_wel_adapter.py
```diff
@@ -143,6 +146,7 @@ def __init__(
         self,
         cellid,
         rate,
+        id,
```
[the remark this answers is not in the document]
JoerivanEngelen: Mf6Wel is not really initiated by users themselves, I refrain from doing this as it can introduce more potential bugs.

imod/mf6/mf6_wel_adapter.py
```diff
@@ -128,6 +128,9 @@ class Mf6Wel(BoundaryCondition, IPackage):
     _pkg_id = "wel"
 
     _period_data = ("cellid", "rate")
+    # Workaround that this needs to be ignored in get_period_varnames. This can
```
ClaireDons: I don't really understand what this workaround is for from this sentence.

imod/msw/model.py
```diff
@@ -302,8 +303,21 @@ def get_pkgkey(
         str
             The key of the package of type ``pkg_type``.
         """
+        # Loop over all packages in the model and match based on filename
```
ClaireDons, line 306: This is the solution we spoke about on Thursday right?

imod/msw/utilities/imod5_converter.py
ClaireDons, line 1: Maybe because I'm still quite unfamiliar with the codebase but some short docstrings would be useful to get more context (like already exists for some methods).

imod/msw/sprinkling.py
```diff
@@ -214,10 +338,249 @@ def from_imod5_data(cls, imod5_data: Imod5DataDict) -> "Sprinkling":
...
+        msw_mf6_sprinkling_df["svat_groundwater"] = _get_svat_groundwater_for_wells(
+            msw_mf6_sprinkling_df, svat_aligned
+        )
+        is_point_inside = msw_mf6_sprinkling_df["svat_groundwater"] > 0
+        # Select columns that need to be written to scap_svat.inp
+        inside_df = msw_mf6_sprinkling_df.loc[
+            is_point_inside, ["svat", "layer", "svat_groundwater"]
+        ]
+        inside_df["svat"] = inside_df["svat"].astype(int)
+        # Set wells with layer > 0 to groundwater abstraction, and wells with layer = 0
```
ClaireDons: What happens to surface water abstraction and if a layer is equal to 0?

imod/tests/test_msw/test_sprinkling.py
```diff
@@ -119,253 +137,439 @@ def test_simple_model_some_svats(fixed_format_parser):
...
+              [0, 2, 0],
+              [0, 3, 0],],
+             [[0, 1, 0],
+              [0, 2, 0],
+              [0, 3, 0]]]
+        )
+        # fmt: on
+        case_data.x_p = [2.0, 2.0, 2.0]
+        case_data.y_p = [3.0, 2.0, 1.0]
+        case_data.layer_p = [1, 2, 3]
```
ClaireDons, line 222: Does the case were the layer is equal to zero need to be tested? (I think that's a possibility?)

imod/msw/sprinkling.py
```diff
@@ -214,10 +338,249 @@ def from_imod5_data(cls, imod5_data: Imod5DataDict) -> "Sprinkling":
...
-        if isinstance(cap_data["artificial_recharge_layer"], pd.DataFrame):
-            data = _sprinkling_data_from_imod5_ipf(cap_data)
-        else:
-            data = _sprinkling_data_from_imod5_grid(cap_data)
+        data = sprinkling_data_from_imod5_grid(cap_data)
 
         return cls(**data)
+
+
+class Sprinkling(SprinklingGrid):
```
ClaireDons, line 358: I hadn't thought of having a deprecated class, but it seems like a good solution to the problem.

imod/tests/test_msw/test_sprinkling.py
ClaireDons, line 1: I enjoyed looking at how you set up the tests, some docstrings for the grid cases (like for the point cases) would be nice to make it easier to follow what the cases are.

ClaireDons (commented):
Nice work, that's quite a significant added functionality! Just a few comments for now, mainly for clarity (for me).

imod/msw/sprinkling.py
```diff
@@ -25,61 +32,127 @@ def _ravel_per_subunit(da: xr.DataArray) -> np.ndarray:
...
+    Create a dataframe of sprinkling well points from the sprinkling dataset and
+    merge it with the mf6_cellid_df to get the row/col of each well.
+    """
+    # Get point data from sprinkling dataset and convert to dataframe for easy merging
+    points_keys = [
+        key for key, da in sprinkling_dataset.data_vars.items() if "id" in da.dims
+    ]
+    sprinkling_points_df = (
+        sprinkling_dataset[points_keys].drop_vars(["dx", "dy"]).to_dataframe()
     )
```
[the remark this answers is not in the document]
JoerivanEngelen: This is wrong: dx and dy are present as coordinates, not as variables. That's why we explicitly need to drop them with (the confusing name) ``drop_vars``. To avoid troubles for users, I'll add a ``errors="ignore"`` option

imod/mf6/mf6_wel_adapter.py
JoerivanEngelen: Changed to: 
```python
    # Add "id" to optional_data classvar as a workaround that this needs to be
    # ignored in get_period_varnames. This can become an _auxiliary_data in the
    # future with some extra work.
```

imod/msw/model.py
JoerivanEngelen: Yes indeed! Instead of doing a typecheck, I'm doing a check for output filename, which is unique.

imod/msw/utilities/imod5_converter.py
JoerivanEngelen: Fair point, I've added docstrings

imod/msw/sprinkling.py
JoerivanEngelen: Thanks :-)

imod/tests/test_msw/test_sprinkling.py
JoerivanEngelen: Very good point! I added some test cases for this, it turned out this required some fixes, and while I was at it, I realized that I could reduce the lengthy _render method a lot, as I was basically conducting the same logic twice in different ways for points outside the svat grid and when layer = 0

imod/tests/test_msw/test_sprinkling.py
JoerivanEngelen: Fair point, I've added docstrings to these cases

ClaireDons (approved):
LGTM, glad the review was helpful!

PR #1886, opened 2026-07-23
Issue #1844 performance idf open subdomains

### LINKED ISSUE #1844: [Bug] Big performance regression on `idf.open_subdomains`
Gijs @gmcmjanssen reported that `open_subdomains` is unusably slow. I checked the git history, the relevant change dates back to Jan/Feb 2024, with the refactoring. https://github.com/Deltares/imod-python/issues/736
The issue contains some performance tests, but they were apparently inadequate.

To help Gijs, I've collected the old function as loose functions, updated with the new locations of some of the utilities. He reports this is fast again.

We should investigate and reproduce the slowdown. The snippet below is the logic pre-2024, so it should be relatively easily to test and benchmark; generate some IDFs (a reasonable number), and try to read and merge them.

I wrote several notes in that issue:

> So the thing to notice here is that open_subdomains is much faster without the compute. This is somewhat expected, it looks at a lot less data, it only opens a single IDF headers of a partition. It's slower during the compute, because it ends up doing slight more work (it calls searchsorted for every time slice).

> In terms of idf.open_subdomains:
>
> It's nice that the current implementation does the least amount of work, but in most cases, the function is used to open all files, merge the data, and write to a new file. So the initial performance boost is not as important, arguably (the penalty is the same as manually calling idf.open on every individual partition).
> It's a lot less flexible, because it doesn't have decent pattern and multi-dimension support. Even if you provide a pattern, it always requires time, layer, subdomain.
There are some additional details: it detects species dimension in the name, and makes sure to cast to int, and then sort. Arguably this should/could be included in the general logic of imod.idf open as well. Currently, you get int species back if you use open_subdomains, but you'd get a str species back if you use idf.open...
Is there a way to get to maintain the current (initial) performance?
>
> The key here is to only look at the names of the files, and open the first IDF of every partition.
> However, you need to adjust a bunch of logic from array_io.reading._load
> Currently, it requires both the filename and the file header to build the DataArray dimensions and coordinates.
> Given the amount of work and potential for new bugs (for IDF support(!)), I don't think it's worth it.
> So my conclusion so far:
> 
> Everything except "new_open_subdomains" is a necessary addition because xr.merge is so slow.
> If we replace the current open_subdomains, we should add a line in imod.util.decompose so that species are always converted  to int (if people have multiple named species in IDFs that cannot be converted to int, they can use a pattern with a different name).
> The new open_subdomains is more flexible, we just at least add test cases with multiple types of chunking (none; time only; time and layer; species, time, and layer).

Importantly, it looks like this assumption is just dead wrong:

> So the initial performance boost is not as important, arguably (the penalty is the same as manually calling idf.open on every individual partition).

Old, quick, version that works with current utilties:

```python
 import collections
import glob
import pathlib
import re

import dask
import numpy as np
import xarray as xr
from imod import util
from imod.formats import array_io
from imod.formats.idf import _read, header
from imod.util.spatial import _xycoords
from imod.util.time import _convert_datetimes


def _merge_subdomains(pathlists, use_cftime, pattern):
    das = []
    for paths in pathlists.values():
        headers = [header(p, pattern) for p in paths]
        das.append(array_io.reading._load(paths, use_cftime, _read, headers))

    x = np.unique(np.concatenate([da.x.values for da in das]))
    y = np.unique(np.concatenate([da.y.values for da in das]))

    nrow = y.size
    ncol = x.size
    nlayer = das[0].coords["layer"].size
    if "species" in das[0].dims:
        has_species = True
        nspecies = das[0].coords["species"].size
        out = np.full((nspecies, 1, nlayer, nrow, ncol), np.nan)
    else:
        has_species = False
        out = np.full((1, nlayer, nrow, ncol), np.nan)

    for da in das:
        ix = np.searchsorted(x, da.x.values[0], side="left")
        iy = nrow - np.searchsorted(y, da.y.values[0], side="right")
        ysize, xsize = da.shape[-2:]
        if has_species:
            out[:, :, :, iy : iy + ysize, ix : ix + xsize] = da.values
        else:
            out[:, :, iy : iy + ysize, ix : ix + xsize] = da.values

    return out


def open_subdomains(path, use_cftime=False, pattern=None):
    """
    Combine IDF files of multiple subdomains.

    Parameters
    ----------
    path : str, Path or list
    use_cftime : bool, optional
    pattern : regex pattern, optional

    Returns
    -------
    xarray.DataArray

    """
    if isinstance(path, pathlib.Path):
        path = str(path)
    paths = glob.glob(path)
    n = len(paths)
    if n == 0:
        raise FileNotFoundError(f"Could not find any files matching {path}")

    if pattern is None:
        pattern = re.compile(
            r"[\w-]+_(?P<time>[0-9-]+)_l(?P<layer>[0-9]+)_p(?P<subdomain>[0-9]{3})",
            re.IGNORECASE,
        )
        pattern_species = re.compile(
            r"[\w-]+_c(?P<species>[0-9]+)_(?P<time>[0-9-]+)_l(?P<layer>[0-9]+)_p(?P<subdomain>[0-9]{3})",
            re.IGNORECASE,
        )
    else:
        pattern_species = None

    # There are no real benefits to itertools.groupby in this case, as there's
    # no benefit to using a (lazy) iterator in this case
    grouped_by_time = collections.defaultdict(
        functools.partial(collections.defaultdict, list)
    )
    count_per_subdomain = collections.defaultdict(int)  # used only for checking counts
    timestrings = []
    layers = []
    numbers = []
    species = []

    # Check if species are present:
    if pattern_species and pattern_species.search(paths[0]) is not None:
        has_species = True
        pattern = pattern_species
    else:
        has_species = False

    for p in paths:
        search = pattern.search(p)
        timestr = search["time"]
        layer = int(search["layer"])
        number = int(search["subdomain"])
        grouped_by_time[timestr][number].append(p)
        count_per_subdomain[number] += 1
        numbers.append(number)
        layers.append(layer)
        timestrings.append(timestr)
        if has_species:
            species.append(int(search["species"]))

    # Test whether subdomains are complete
    numbers = sorted(set(numbers))
    first = numbers[0]
    first_len = count_per_subdomain[first]
    for number in numbers:
        group_len = count_per_subdomain[number]
        if group_len != first_len:
            raise ValueError(
                f"The number of IDFs are not identical for every subdomain. "
                f"Subdomain p{first} has {first_len} IDF files, subdomain p{number} "
                f"has {group_len} IDF files."
            )

    if has_species:
        pattern = r"{name}_c{species}_{time}_l{layer}_p\d+"
    else:
        pattern = r"{name}_{time}_l{layer}_p\d+"
    timestrings = list(grouped_by_time.keys())

    # Prepare output coordinates
    coords = {}
    first_time = timestrings[0]
    samplingpaths = [
        pathlist[first] for pathlist in grouped_by_time[first_time].values()
    ]
    headers = [header(path, pattern) for path in samplingpaths]
    subdomain_bounds = [(h["xmin"], h["xmax"], h["ymin"], h["ymax"]) for h in headers]
    subdomain_cellsizes = [(h["dx"], h["dy"]) for h in headers]
    subdomain_coords = [
        _xycoords(bounds, cellsizes)
        for bounds, cellsizes in zip(subdomain_bounds, subdomain_cellsizes)
    ]
    coords["y"] = np.unique(
        np.concatenate([coords["y"] for coords in subdomain_coords])
    )[::-1]
    coords["x"] = np.unique(
        np.concatenate([coords["x"] for coords in subdomain_coords])
    )
    coords["layer"] = np.array(sorted(set(layers)))

    times = [util.to_datetime(timestr) for timestr in timestrings]
    times, use_cftime = _convert_datetimes(times, use_cftime)
    if use_cftime:
        # unique also sorts
        coords["time"] = xr.CFTimeIndex(np.unique(times))
    else:
        coords["time"] = np.unique(times)

    if has_species:
        coords["species"] = np.array(sorted(set(species)))
        shape = (
            coords["species"].size,
            1,
            coords["layer"].size,
            coords["y"].size,
            coords["x"].size,
        )
        dims = ("species", "time", "layer", "y", "x")
        time_axis = 1
    else:
        shape = (1, coords["layer"].size, coords["y"].size, coords["x"].size)
        dims = ("time", "layer", "y", "x")
        time_axis = 0

    # Collect and merge data
    merged = []
    dtype = headers[0]["dtype"]
    sorted_order = np.argsort(times)  # get time ordering right before merging
    sorted_timestrings = np.array(timestrings)[sorted_order]
    for timestr in sorted_timestrings:
        group = grouped_by_time[timestr]
        # Build a single array per timestep
        timestep_data = dask.delayed(_merge_subdomains)(group, use_cftime, pattern)
        dask_array = dask.array.from_delayed(timestep_data, shape, dtype=dtype)
        merged.append(dask_array)
    data = dask.array.concatenate(merged, axis=time_axis)

    # Get tops and bottoms if possible
    headers = [header(path, pattern) for path in grouped_by_time[first_time][first]]
    tops = [c.get("top", None) for c in headers]
    bots = [c.get("bot", None) for c in headers]
    layers = [c.get("layer", None) for c in headers]
    _, unique_indices = np.unique(layers, return_index=True)
    all_have_z = all(map(lambda v: v is not None, itertools.chain(tops, bots)))
    if all_have_z:
        if coords["layer"].size > 1:
            coords = array_io.reading._array_z_coord(coords, tops, bots, unique_indices)
        else:
            coords = array_io.reading._scalar_z_coord(coords, tops, bots)

    return xr.DataArray(data, coords, dims)
 ```

### PR DESCRIPTION
**Pull request by JoerivanEngelen**
Fixes #1844

# Description
Reduce dask task graph considerably by only chunking per time (and optionally per species). To avoid having to construct the coords manually again from the headers like the previous implementation, we eagerly call merge_partitions on the first timestep to just get the coords (but do not load them into memory). This implementation has similar performance to the original code. I had to modify the return statements in ``imod.typing.structured.py`` somewhat to avoid unnecessary creation of xr.DataArrays and xr.Datasets. This might look a bit confusing at first, I added some docstrings to clarify.

**Other things I tried:**

I also tried rechunking to layers after calling merge_partitions, so that not so much code had to be changed, but this still resulted in poor performance (5 minutes versus 6 seconds with 37400 idfs) as dask still had to optimize the task graph. (Still better than the 30+ minutes it cost before)

TODOs:
- [x] Make time dimension optional
- [x] See if grouping dicts twice is necessary
- [x] Add a user acceptance test to deal with many IDFs, hopefully teamcity can handle this.

# Checklist

- [x] Links to correct issue
- [x] Unit tests were added

### CODE REVIEW

**Comment by Huite** [comment]
Some notes:

* There may be slight overhead in constructing DataArrays in the `merge_subdomains` function; this requires alignment checks etc.
* This will only show up in a `.compute` operation.
* If relevant, a workaround is to essentially define/modify an internal functions which returns three arguments (data, dims, coords) rather than a DataArray.
* The obviates the `.values` wrapper here -- eliminating that might be nice anyway!
* The template can still be constructed using the higher level function which does return a DataArray: the validation xarray performs is worthwhile.

EDIT: overhead is minimal; for 3000 time steps with 50 layers, 1000 by 1000 rows & columns, it's about 1 ms of overhead.
So the only reason to do so is to avoid the .values wrapper.

**Comment by JoerivanEngelen** [comment]
Update: I managed to further improve performance by a factor 2, so that it is equal to the original solution by avoiding excessive unnecessary casting to ``xr.DataArray``. It complicates the code somewhat, as there is now a separate function to return the components (data, coords, dims) separately, as well as a function which wraps this into a ``xr.DataArray``.

File: imod/formats/idf.py [hunk 1]
```diff
@@ -243,6 +314,11 @@ def open_subdomains(
     xarray.DataArray
 
     """
+    # This function is a wrapper around open() that groups by subdomain and
```
ClaireDons, line 317: Is there a reason you didn't want this (or part of this) to be part of the docstrings?

imod/formats/idf.py
```diff
@@ -254,30 +330,70 @@ def open_subdomains(
...
+        grouped_by_time[time_key][match["subdomain"]].append(p)
+
+    # Sort and convert times before calling _merge_subdomains so that
+    # use_cftime is already correct when the template is built.
+    raw_times_sorted = sorted(grouped_by_time.keys())
+    converted_times, use_cftime = imod.util.time._convert_datetimes(
+        raw_times_sorted, use_cftime
+    )
+
+    # Call _merge_subdomains eagerly for the first timestep to obtain a
```
ClaireDons, line 359: What happens if there are inconsistencies in the timesteps? i.e if further down the line there is data missing or something or are these files always consistent?

ClaireDons (commented):
Nice, it was quite interesting to go through the changes. I just have a couple of questions but otherwise looks good to me :)

imod/formats/idf.py
JoerivanEngelen: The amount of files is checked in ``check_subdomain_consistency`` so that catches part of the potential errors here. For example, if subdomain 1 has 10 timesteps and subdomain 2 has 11 timesteps, this will be caught. ``imod.idf.open_subdomains`` is mostly used to post-process iMOD-WQ output files, which are consistent. I don't think this will be used to process user-generated IDFs, which of course is more error prone. If the dataset is inconsistent, xarray will probably throw an error somewhere.

imod/formats/idf.py
JoerivanEngelen: Yes, this function is public API, so the docstring is written for users, this comment is intended for developers.

imod/formats/idf.py
ClaireDons: Yea it's probably fine for now

ClaireDons (approved):

PR #1887, opened 2026-07-23
Issue 1882 - Fix rolling out of character limit update

### LINKED ISSUE #1882: [Bug] - Character limit change not fully rolled out
**Bug description**
[I saw our weekly test broke](https://dpcbuild.deltares.nl/buildConfiguration/iMOD6_IMODPython_Windows_AcceptanceTests/7569143?expandedTest=build%3A%28id%3A7569143%29%2Cid%3A2000000000), this is because we forgot about the ``imod.mf6.Modflow6Model.__setitem__`` method, which also checks the amount of characters of a key.

**Steps to reproduce**
1. ``KeyError: "Received key with more than 16 characters: 'wel-1-WEL_OVERIG_NON_NHI'Modflow 6 has a character limit of 16."``
2. Can you create a minimum reproducible example?
    ```
    model = imod.mf6.Modflow6Model()
    model["a_name_longer_than_16_char_but_not_40"] = {}
    ```

Additionally, the truncation of the wells in ``imod.mf6.GroundwaterFlowModel.from_imod5_data`` should be done AFTER the numbers are added.

ClaireDons:
Fixes #1882 

# Description
We now only raise an error if a key is longer than 40 characters and updated the relevant tests that were still looking for 16. After checking, we already truncate the names after the numbers are added (so the logic was already correct). 

# Checklist

- [x] Links to correct issue
- [ ] Unit tests were added

### DIFF
--- a/docs/api/changelog.rst
+++ b/docs/api/changelog.rst
@@ -93,6 +93,8 @@ Fixed
 Changed
 ~~~~~~~
 
+- Increased the character limit to 40 in :class:`imod.mf6.Modflow6Model` for all 
+  keys assigned to a Modflow6 model.
 - ``proportion_depth`` and ``proportion_rate`` in
   :class:`imod.mf6.Evapotranspiration` are now optional variables. If provided,
   now require ``"segment"`` dimension when ``proportion_depth`` and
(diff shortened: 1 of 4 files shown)


JoerivanEngelen (approved):
Looks good, could you update the changelog? That we increased the max amount of characters to 40? I see there is a specific comment under ``fixed`` for imod5 with wells specific, but the change we make here is more general, as it applies to all keys assigned to a Modflow6 model.

Approving in advance

PR #1889, opened 2026-07-27
Issue 1820 create user guide example for logging

### LINKED ISSUE #1820: [DOCUMENTATION] - Add section to user guide and "How do I" about logging
We currently lack documentation on logging for regular users. [There is this](https://deltares.github.io/imod-python/api/generated/logging/imod.logging.html#module-imod.logging), but that is very oriented to developers.

ClaireDons:
Fixes #1820 

# Description
Add a section to the user guide about logging, including how to configure the type of logger, logging level and writing to both the default or a custom file. 

# Checklist

- [x] Links to correct issue
- [ ] Unit tests were added


examples/user-guide/13-logging.py
```diff
@@ -0,0 +1,82 @@
+"""
+Logging in iMOD-Python
+======================================
+
+iMod-python supports logging through both the standard
```
JoerivanEngelen: Correct writing of the package in text is:

```suggestion
iMOD Python supports logging through both the standard
```

examples/user-guide/13-logging.py
```diff
@@ -0,0 +1,82 @@
...
+# simulations, to keep logging output organised for debugging and verification.
+# Here, we also set `add_default_stream_handler=False` to avoid logging to the console.
+# By default, this is usually True, meaning logging is also sent to the console.
+
+from contextlib import redirect_stdout
+
+logfile_path = "open_simulation.log"
+
+with open(logfile_path, "w") as f:
+    with redirect_stdout(f):
```
JoerivanEngelen, line 77: What is this for?

JoerivanEngelen (approved):
Looks good, I have some remarks. Approving in advance.

examples/user-guide/13-logging.py
ClaireDons: It just redirects the output to the file and is fairly straightforward to implement. I'll add a comment to make it clearer.
It's instead of using `with open(logfile_path, "w") as sys.stdout:`, although I could do that too if you prefer, it would just be a few more lines of code. It would make it a little harder to read for a user possibly?

JoerivanEngelen: I'm not really fan of the double with statement here, it makes things look a lot more complex. I agree that ``with open(logfile_path, "w") as sys.stdout`` is somewhat obfuscating, but I slightly prefer it over the double with statement.

ClaireDons: After trying things and talking more about it we decided that using the context manager with redirect_stdout() was a little more robust when re-running notebook cells and handles automatically cleaning things up after the block is run.

JoerivanEngelen (approved):

JoerivanEngelen:
I see two jobs failing, but I think that is a TeamCity issue.

PR #1901, opened 2026-08-05
Fail the build when one of the commands in the script fail

### LINKED ISSUE #1900: The "Pixi update" pipeline reports succeeded even though its substeps fails

Manangka:
Fixes #1900

# Description
The pixi update pipeline doesnt handle errors produced by any of the commands called in the script correctly.
Instead it just continue with the next lines of the script which results in the pipeline being green while actually it failed.

In this PR the same fix is applied to imod-python as has been done in imod_coupler.
https://github.com/Deltares/imod_coupler/pull/495

As a result an error is now handled correctly:
https://dpcbuild.deltares.nl/buildConfiguration/iMOD6_Coupler_UpdateDependencies/7684833?buildTab=log&linesState=524&logView=flowAware

image
 

# Checklist

- [x] Links to correct issue
- [ ] Unit tests were added


JoerivanEngelen (approved):
LGTM!

PR #1912, opened 2026-08-31
Issue #1821: Improved performance for allocation & sparse/empty layer handling.

### LINKED ISSUE #1821: [FEATURE] - Allocate RIV, DRN, RCH: drop unused layers
We now always broadcast data to all layers in our allocation routines, whereas this shouldn't be necessary. This will make processing data unnecessary slow, especially when dealing with models with lots of layers and many timesteps. 

I'm not sure if the splitting, clipping, regridding, and masking methods silently broadcast data to all layers, so that is something which needs additional testing.

LuukBlom:
Fixes #1821

# Description

- Add `drop_empty_layers: bool = False` to `allocate_<riv/drn/rch>_cells` functions in `imod.prepare.topsystem.allocation` to remove fully empty layers from the grids to prevent passing them along to reprojection/regridding operations.
- Improved imod.prepare.layerregrid._regrid_layers to precompute valid layer indices once per column, instead of re-checking isnan for every (ii, jj) pair.

# Default value discussion

I chose the default for `drop_empty_layers` to be `False` to have it as a backwards compatible opt-in new behaviour. But perhaps the default should be True, since it is somewhat difficult for users to actually use the opt-in behaviour. (its an internal function call, so we might need to add it to some other functions as an arg with default as well?)

Setting it to `True` broke some tests (which is expected since these tests asserted on the shape of the returned objects).
I think setting the default to `True` and updating those tests to expect the new format is the way to go, but I'd like to discuss with @JoerivanEngelen before doing that. For now, I have left it at False to not break the existing tests, and added tests that call this path with `drop_empty_layers=True`.

Alternative solution could be that we set the default to True and call a function similar to `reindex_to_full_layers` (see usage in `test_topsystem_layer_preservation.py`) in places where its required to have all layers (even empty ones)

# Checklist

- [x] Links to correct issue
- [x] Unit tests were added
- [x] **If pixi.lock was changed**: Ran `pixi run generate-sbom` and committed changes

---

## Investigation: cost of broadcasting to all layers in allocation/regrid pipeline

<details>

Spent some time to investigate this issue to find 1) the slow/expensive part of the code and 2) whether and how to optimize.

### Summary

The allocation boolean logic itself (`allocate_riv_cells`, `allocate_drn_cells`, etc.) is **not** the bottleneck, it's cheap regardless of layer count. 
The real cost is in `LayerRegridder._regrid_layers`, which has a nested `n_layers_dst × n_layers_src` loop per (row, col). When a package's real data only occupies a handful of layers but is broadcast/kept at full model-layer size before regridding, cost scales roughly **quadratically with total layer count**, not with the number of layers that actually contain data.

Measured: **~8.4s** at 300 layers vs **~1ms** at 2 layers, for a package whose real data lives in only 2 layers throughout.

### Where the full-layer array comes from

`imod/prepare/topsystem/allocation.py` (`_allocate_cells__stage_to_riv_bot`, `_allocate_cells__at_elevation`, `_allocate_cells__first_active_to_elevation`, etc.) all compute their result by comparing `stage`/`bottom_elevation` (planar, no layer dim, enforced by `PLANAR_GRID.validate`) against `top`/`bottom`/`active`, which always carry the **full** model layer coordinate:

```python
is_above_lower_bound = get_above_lower_bound(bottom_elevation, top_layered)
is_below_upper_bound = stage >= bottom
riv_cells = is_below_upper_bound & is_above_lower_bound
```

This is not changeable, we can't know in advance which layers a planar stage/elevation grid intersects without comparing against every layer's top/bottom. That comparison is a single vectorized boolean op, and is cheap at any layer count (confirmed below).

The problem is that `riv_cells` (and the package's `stage`/`conductance`/etc. once built via `.where(riv_cells)`) then keeps the **full `n_layers` coordinate**, mostly `False`/`NaN`, and nothing downstream trims it back down before it's handed to regridding, clipping, masking, or splitting.

### Where the cost actually is: `LayerRegridder._regrid_layers`

From `imod/prepare/regrid_layers.py` (numba-jitted):

```python
for ii in range(nlayer_dst):
    ...
    for jj in range(nlayer_src):
        st = src_t[jj]
        sb = src_b[jj]
        if np.isnan(st) or np.isnan(sb):
            continue
        ...
```

This is a nested loop over **`nlayer_dst × nlayer_src × nrow × ncol`**. Even though the `isnan` check that skips empty layers is cheap per-iteration, the loop still has to *visit* every `(dst_layer, src_layer)` pair for every cell, so an all-NaN source layer still costs an iteration, for every destination layer. If a package's real data occupies only 2 of N layers, this loop still costs `nlayer_dst × N` iterations instead of `nlayer_dst × 2`.

Since `nlayer_dst` also scales with total model layers in typical regridding (source and destination models usually share vertical discretization), the practical cost scales roughly **quadratically** with total layer count, independent of how many layers actually contain real data.

### Reproducer

<details>

```python
import time

import numpy as np
import xarray as xr

from imod.prepare import LayerRegridder


def make_layered(n_layers, nrow, ncol, real_layers, trim=False):
    x = np.arange(ncol) * 10.0
    y = np.arange(nrow) * -10.0
    layer = np.arange(1, n_layers + 1)

    top = np.zeros((n_layers, nrow, ncol))
    bottom = np.zeros((n_layers, nrow, ncol))
    for k in range(n_layers):
        top[k] = -k
        bottom[k] = -k - 1

    source = np.full((n_layers, nrow, ncol), np.nan)
    source[:real_layers] = 1.0

    coords = {"layer": layer, "y": y, "x": x}
    dims = ("layer", "y", "x")
    src_da = xr.DataArray(source, coords, dims)
    top_da = xr.DataArray(top, coords, dims)
    bot_da = xr.DataArray(bottom, coords, dims)

    if trim:
        # This is what allocation SHOULD hand to the regridder once
        # trimmed: only the layers that actually contain data.
        src_da = src_da.isel(layer=slice(0, real_layers))
        top_da = top_da.isel(layer=slice(0, real_layers))
        bot_da = bot_da.isel(layer=slice(0, real_layers))

    return src_da, top_da, bot_da


def time_regrid(
    n_layers, real_layers=2, nrow=200, ncol=200, trim_source=False, repeats=3
):
    src, src_top, src_bot = make_layered(
        n_layers, nrow, ncol, real_layers, trim=trim_source
    )

    # Destination always spans the full model layer range - that part
    # is unavoidable, since regridding must produce a value (or nan)
    # for every destination layer.
    dst_layer = np.arange(1, n_layers + 1)
    full_top, _, _ = make_layered(n_layers, nrow, ncol, real_layers, trim=False)
    dst_top = (
        full_top.rename({"layer": "layer"}).assign_coords(layer=dst_layer) * 0
        + full_top.values
    )
    dst_top = xr.DataArray(
        np.stack([-np.full((nrow, ncol), float(k)) for k in range(n_layers)]),
        {"layer": dst_layer, "y": src.y, "x": src.x},
        ("layer", "y", "x"),
    )
    dst_bot = dst_top - 1.0

    regridder = LayerRegridder(method="mean")
    regridder.regrid(src, src_top, src_bot, dst_top, dst_bot)  # warm up JIT

    times = []
    for _ in range(repeats):
        t0 = time.perf_counter()
        regridder.regrid(src, src_top, src_bot, dst_top, dst_bot)
        times.append(time.perf_counter() - t0)

    label = "TRIMMED src" if trim_source else "dense src (padded w/ nan)"
    print(
        f"n_layers={n_layers:4d} real_layers={real_layers}  [{label}]  best={min(times):.4f}s"
    )


if __name__ == "__main__":
    print("-- dense (current behaviour): source kept at full n_layers, nan-padded --")
    for n in [2, 5, 10, 30, 100, 300]:
        time_regrid(n_layers=n, real_layers=2, trim_source=False)

    print("\n-- trimmed: source pre-cut to its 2 real layers before regrid --")
    for n in [2, 5, 10, 30, 100, 300]:
        time_regrid(n_layers=n, real_layers=2, trim_source=True)

```

</details>

### Results

```
-- dense (current behaviour): source kept at full n_layers, nan-padded --
n_layers=   2 real_layers=2  [dense src (padded w/ nan)]  best=0.0051s
n_layers=   5 real_layers=2  [dense src (padded w/ nan)]  best=0.0070s
n_layers=  10 real_layers=2  [dense src (padded w/ nan)]  best=0.0143s
n_layers=  30 real_layers=2  [dense src (padded w/ nan)]  best=0.0985s
n_layers= 100 real_layers=2  [dense src (padded w/ nan)]  best=1.0968s
n_layers= 300 real_layers=2  [dense src (padded w/ nan)]  best=9.7136s

-- trimmed: source pre-cut to its 2 real layers before regrid --
n_layers=   2 real_layers=2  [TRIMMED src]  best=0.0041s
n_layers=   5 real_layers=2  [TRIMMED src]  best=0.0041s
n_layers=  10 real_layers=2  [TRIMMED src]  best=0.0051s
n_layers=  30 real_layers=2  [TRIMMED src]  best=0.0106s
n_layers= 100 real_layers=2  [TRIMMED src]  best=0.0342s
n_layers= 300 real_layers=2  [TRIMMED src]  best=0.1197s
```

In every case, only 2 of `n_layers_src` layers ever contain real (non-NaN) data. Yet runtime grows from 1.4ms to 9.7s (~6000x) purely as a function of total layer count. This confirms the cost scales with the model's total layer count, not with the amount of actual topsystem data, and that `LayerRegridder` is the concrete hot spot to fix.

Trimming the empty layers before regridding is a good approach for performance.

</details>

### DIFF
--- a/imod/prepare/layerregrid.py
+++ b/imod/prepare/layerregrid.py
@@ -30,10 +30,7 @@ def _valid_layer_indices(top_col, bot_col, out):
 
 @numba.njit(cache=True)
 def _regrid_layers(src, dst, src_top, dst_top, src_bot, dst_bot, method):
-    """
-    Maps one set of layers onto the other, skipping all-nan layers up front
-    per column instead of checking nan inside the nested layer loop.
-    """
+    """Maps one set of layers onto the other."""
     nlayer_src, nrow, ncol = src.shape
     nlayer_dst = dst.shape[0]
 
--- a/imod/prepare/layerregrid.py
+++ b/imod/prepare/layerregrid.py
@@ -12,14 +12,35 @@ METHODS.pop("multilinear")
 
 
 @numba.njit(cache=True)
+def _valid_layer_indices(top_col, bot_col, out):
+    """
+    Fill `out` (int64 array, same length as top_col) with the indices of
+    layers that have non-nan top AND bottom, at a single (row, col) column.
+    Returns the count of valid entries. Avoids allocating a new array per
+    column call.
+    """
+    count = 0
+    n = top_col.shape[0]
+    for k in range(n):
+        if not (np.isnan(top_col[k]) or np.isnan(bot_col[k])):
+            out[count] = k
+            count += 1
+    return count
+
+
+@numba.njit(cache=True)
 def _regrid_layers(src, dst, src_top, dst_top, src_bot, dst_bot, method):
     """
-    Maps one set of layers unto the other.
+    Maps one set of layers onto the other, skipping all-nan layers up front
+    per column instead of checking nan inside the nested layer loop.
     """
     nlayer_src, nrow, ncol = src.shape
     nlayer_dst = dst.shape[0]
+
     values = np.zeros(nlayer_src)
     weights = np.zeros(nlayer_src)
+    src_valid_idx = np.empty(nlayer_src, dtype=np.int64)
+    dst_valid_idx = np.empty(nlayer_dst, dtype=np.int64)
 
     for i in range(nrow):
         for j in range(ncol):
@@ -28,22 +49,26 @@ def _regrid_layers(src, dst, src_top, dst_top, src_bot, dst_bot, method):
             src_b = src_bot[:, i, j]
             dst_b = dst_bot[:, i, j]
 
-            # ii is index of dst
-            for ii in range(nlayer_dst):
+            # Precompute valid layer indices ONCE per column, instead of
+            # re-checking isnan for every (ii, jj) pair.
+            n_src_valid = _valid_layer_indices(src_t, src_b, src_valid_idx)
+            if n_src_valid == 0:
+                continue
+            n_dst_valid = _valid_layer_indices(dst_t, dst_b, dst_valid_idx)
+            if n_dst_valid == 0:
+                continue
+
+            for di in range(n_dst_valid):
+                ii = dst_valid_idx[di]
                 dt = dst_t[ii]
                 db = dst_b[ii]
-                if np.isnan(dt) or np.isnan(db):
-                    continue
 
                 count = 0
                 has_value = False
-                # jj is index of src
-                for jj in range(nlayer_src):
+                for sj in range(n_src_valid):
+                    jj = src_valid_idx[sj]
                     st = src_t[jj]
                     sb = src_b[jj]
-
-                    if np.isnan(st) or np.isnan(sb):
-                        continue
 
                     overlap = common._overlap((db, dt), (sb, st))
                     if overlap == 0:
@@ -53,12 +78,11 @@ def _regrid_layers(src, dst, src_top, dst_top, src_bot, dst_bot, method):
                     values[count] = src[jj, i, j]
                     weights[count] = overlap
                     count += 1
-                else:
-                    if has_value:
-                        dst[ii, i, j] = method(values, weights)
-                        # Reset
-                        values[:count] = 0
-                        weights[:count] = 0
+
+                if has_value:
+                    dst[ii, i, j] = method(values, weights)
+                    values[:count] = 0
+                    weights[:count] = 0
 
     return dst
 
--- a/imod/prepare/topsystem/allocation.py
+++ b/imod/prepare/topsystem/allocation.py
@@ -581,9 +581,7 @@ def _allocate_cells__at_first_active(
     return topsystem_upper_active, None
 
 
-def _drop_empty_layers(
-    grid: GridDataArray, spatial_dims: tuple[str, ...] = ("y", "x")
-) -> GridDataArray:
+def _drop_empty_layers(grid: GridDataArray) -> GridDataArray:
     """
     Drop layers that contain no True/non-nan values in any spatial cell
     (and, if present, at any timestep). Keeps the `layer` coordinate but
@@ -597,10 +595,6 @@ def _drop_empty_layers(
     grid: GridDataArray
         Array with a "layer" dimension, typically the output of one of the
         ``_allocate_cells__*`` functions.
-    spatial_dims: tuple[str, ...]
-        Dimensions to reduce over when checking "is this layer used
-        anywhere". Does not include "layer" or "time" by design - a layer
-        used at any timestep, anywhere in the domain, is kept.
 
     Returns
     -------
--- a/imod/prepare/topsystem/allocation.py
+++ b/imod/prepare/topsystem/allocation.py
@@ -71,6 +71,7 @@ def allocate_riv_cells(
     bottom: GridDataArray,
     stage: GridDataArray,
     bottom_elevation: GridDataArray,
+    drop_empty_layers: bool = False,
 ) -> tuple[GridDataArray, Optional[GridDataArray]]:
     """
     Allocate river cells from a planar grid across the vertical dimension.
@@ -96,6 +97,14 @@ def allocate_riv_cells(
     bottom_elevation: DataArray | UgridDatarray
         Planar grid containing river bottom elevations. Is not allowed to have a
         layer dimension.
+    drop_empty_layers: bool, default False
+        If True, drop layers from the result that contain no allocated
+        cells anywhere in the domain. This avoids carrying the package's
+        arrays at full model-layer size through downstream regridding,
+        clipping, masking, and splitting, which can otherwise become
+        expensive for models with many layers relative to how many
+        layers the topsystem package actually occupies. Set to False to
+        keep the previous full-layer-coordinate behaviour.
 
     Returns
     -------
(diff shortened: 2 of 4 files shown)


LuukBlom:
Tests are green locally when calling `pixi run tests`.

There seems to be something wrong with docker + Teamcity though.

Which is very strange as I didnt change anything there.
Re-running the failed checks doesnt fix it unfortunately.

All failures are caused by the same error:

```text
pull containers.deltares.nl/hydrology_product_line_imod/windows-pixi:v0.69.0
  failed to register layer: strconv.ParseInt: parsing "": invalid syntax
Unable to find image 'containers.deltares.nl/hydrology_product_line_imod/windows-pixi:v0.69.0' locally
v0.69.0: Pulling from hydrology_product_line_imod/windows-pixi
fbc33898c8ff: Pulling fs layer
1e22643c3eaa: Pulling fs layer
0318f4e42c5d: Pulling fs layer
e778778b905f: Pulling fs layer
9ffbd341cca3: Pulling fs layer
c6aee63eee57: Pulling fs layer
b88504fedbea: Pulling fs layer
a29a128dd81c: Pulling fs layer
ed50fbf4d281: Pulling fs layer
85bee0e55f42: Pulling fs layer
92ea28bcf781: Pulling fs layer
4e87dcc2f18a: Pulling fs layer
c6aee63eee57: Waiting
b88504fedbea: Waiting
e778778b905f: Waiting
a29a128dd81c: Waiting
92ea28bcf781: Waiting
ed50fbf4d281: Waiting
4e87dcc2f18a: Waiting
85bee0e55f42: Waiting
9ffbd341cca3: Waiting
0318f4e42c5d: Verifying Checksum
0318f4e42c5d: Download complete
e778778b905f: Verifying Checksum
e778778b905f: Download complete
9ffbd341cca3: Verifying Checksum
9ffbd341cca3: Download complete
c6aee63eee57: Verifying Checksum
c6aee63eee57: Download complete
b88504fedbea: Verifying Checksum
b88504fedbea: Download complete
a29a128dd81c: Verifying Checksum
a29a128dd81c: Download complete
ed50fbf4d281: Verifying Checksum
ed50fbf4d281: Download complete
85bee0e55f42: Verifying Checksum
85bee0e55f42: Download complete
92ea28bcf781: Verifying Checksum
92ea28bcf781: Download complete
4e87dcc2f18a: Verifying Checksum
4e87dcc2f18a: Download complete
1e22643c3eaa: Verifying Checksum
1e22643c3eaa: Download complete
fbc33898c8ff: Verifying Checksum
fbc33898c8ff: Download complete
docker: failed to register layer: strconv.ParseInt: parsing "": invalid syntax
Run 'docker run --help' for more information
Process exited with code 125
```

Not really sure what to do about this

LuukBlom:
Spoke to the Devops team and they are aware of the flaky docker problem.
Is an issue thats popped up today and they are working on a fix. 
ETA is unclear, but they will let us know when there are updates

LuukBlom:
TODO proper run of all tests (including user acceptance / weekly) to make sure that this doesnt break anything when set to `True`

We need to decide on whether the default will be True or False, or not configurable and always drop empty layers.
If we decide on `True` or dropping, we also need to update the assertions for various unittests, and perhaps other tests

imod/prepare/topsystem/allocation.py
```diff
@@ -537,3 +579,48 @@ def _allocate_cells__at_first_active(
...
+    else:
+        has_data_per_layer = (~grid.isnull()).any(dim=reduce_dims)
+
+    # Force to plain numpy/bool to avoid triggering a dask compute deep
+    # inside indexing logic more than once.
+    has_data_per_layer = (
+        has_data_per_layer.compute()
+        if hasattr(has_data_per_layer, "compute")
+        else has_data_per_layer
+    )
```
JoerivanEngelen, line 620: Probably good to call a compute here, but the check for hasattr compute is not necessary, ``xr.DataArray.compute()`` is always available and always returns the same thing.


```suggestion
    # Force to plain numpy/bool to avoid triggering a dask compute deep
    # inside indexing logic more than once.
    has_data_per_layer = has_data_per_layer.compute()
```

imod/prepare/topsystem/allocation.py
```diff
@@ -537,3 +579,48 @@ def _allocate_cells__at_first_active(
...
+    """
+    if "layer" not in grid.dims:
+        return grid
+
+    reduce_dims = [d for d in grid.dims if d != "layer"]
+
+    if grid.dtype == bool:
+        has_data_per_layer = grid.any(dim=reduce_dims)
+    else:
+        has_data_per_layer = (~grid.isnull()).any(dim=reduce_dims)
```
JoerivanEngelen, line 612: I think grid.dtype should always be ``bool`` here, so the if else statement should not be necessary here, maybe throw an error if not the case? I might be missing something though.

imod/prepare/layerregrid.py
JoerivanEngelen, line 1: Thanks for optimizing and improving this. Unfortunately, this is not the regridder we are using commonly. We are relying on the stuff under ``imod/common/utilities/regrid``. This is a specific older regridder for which xugrid doesn't have a stand-in replacement (yet). I thing these optimizations are not harmful though

JoerivanEngelen (commented):
The changeset looks good, my plan was indeed to make the default of dropping layers True, so I'm fine with updating the test fixtures/test cases, except for the tests under tests/test_prepare/test_topsystem.py (or change the test and its cases with to test for a list of layer numbers instead of a list with True/False). Testing against the user acceptance tests is a must.

The performance investigation went a bit on the wrong track: ``LayerRegridder`` is not the regridder that is called when regridding MODFLOW 6 simulations. I was interested if feeding this into ``imod.mf6.Modflow6Simulation.regrid_like`` lead to any problems. I guess we'll run soon enough into this if you change the default of ``drop_empty_layers`` to True.

JoerivanEngelen:
Pseudocode to extend tests for allocation:

```python
@parametrize_with_cases(
    argnames="active,top,bottom,stage,bottom_elevation",
    prefix="riv_",
)
@parametrize_with_cases(
    argnames="option,expected_riv,expected_drn", prefix="allocation_", has_tag="riv"
)
def test_riv_allocation(
    active, top, bottom, stage, bottom_elevation, option, expected_riv, expected_drn
):
    actual_riv_da, actual_drn_da = allocate_riv_cells(
        option, active, top, bottom, stage, bottom_elevation
    )

    actual_riv = take_nth_layer_column(actual_riv_da, 0)
    empty_riv = take_nth_layer_column(actual_riv_da, 1)

    if actual_drn_da is None:
        actual_drn = None
        empty_drn = None
    else:
        actual_drn = take_nth_layer_column(actual_drn_da, 0)
        empty_drn = take_nth_layer_column(actual_drn_da, 1)

    np.testing.assert_equal(actual_riv, expected_riv)
    np.testing.assert_equal(actual_drn, expected_drn)
    assert np.all(~empty_riv)
    if empty_drn is not None:
        assert np.all(~empty_drn)

    expected_layers = np.nonzero(expected_riv) + 1

    actual_riv_da, actual_drn_da = allocate_riv_cells(
        option, active, top, bottom, stage, bottom_elevation, drop_empty_layers=True
    )
    assert np.array_equal(actual_riv_da.coords["layer"].values, expected_layers)

```

